PR #86, opened 2021-06-16
Publish in MavenCentral

gmerinojimenez:
Publish releases in MavenCentral.

The steps to publish are going to be:
1. Generate a release with the github action.
2. The result will be a release draft, the user will need to go there and publish the draft.
3. At that moment, the artifact, which is in a staging state in mavenCentral will be published.

There are two github actions, one to publish to staging and the second one to promote the staging repository.

The first one: release.yml sends the library and the catalog to mavenCentral.
And then the second one, closes & promotes the currently opened staging repository

- [x] [The publishing](https://github.com/Telefonica/mistica-android/actions/runs/943082573) 
- [ ] The promoting can't be tested now. I've tested it locally and will do a final test when this is integrated.

### DIFF
--- a/.github/workflows/promote_release.yml
+++ b/.github/workflows/promote_release.yml
@@ -11,6 +11,9 @@ jobs:
 
       - name: Close and promote staging repository
         uses: docker://docker.tuenti.io/android/mistica_builder:2
+        env:
+          MOBILE_MAVENCENTRAL_USER: ${{ secrets.MOBILE_MAVENCENTRAL_USER }}
+          MOBILE_MAVENCENTRAL_PASSWORD: ${{ secrets.MOBILE_MAVENCENTRAL_PASSWORD }}
         with:
           args: './gradlew closeAndReleaseRepository -DSTAGING_PROFILE_ID="$(./gradlew getStagingProfile  | sed -n -e "s/^Received staging profile id: //p")"'
 
--- a/.github/workflows/promote_release.yml
+++ b/.github/workflows/promote_release.yml
@@ -6,5 +6,11 @@ jobs:
   promote:
     runs-on: self-hosted-novum
     steps:
-      - run: 'echo TODO Promote'
-        shell: bash
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Close and promote staging repository
+        uses: docker://docker.tuenti.io/android/mistica_builder:2
+        with:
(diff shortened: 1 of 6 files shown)


.github/workflows/promote_release.yml
```diff
@@ -6,5 +6,14 @@ jobs:
...
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Close and promote staging repository
+        uses: docker://docker.tuenti.io/android/mistica_builder:2
+        env:
+          MOBILE_MAVENCENTRAL_USER: ${{ secrets.MOBILE_MAVENCENTRAL_USER }}
+          MOBILE_MAVENCENTRAL_PASSWORD: ${{ secrets.MOBILE_MAVENCENTRAL_PASSWORD }}
+        with:
+          args: './gradlew closeAndReleaseRepository -DSTAGING_PROFILE_ID="$(./gradlew getStagingProfile  | sed -n -e "s/^Received staging profile id: //p")"'
```
gmerinojimenez, line 18: little trick to get the staging profile and then pass it to closeAndReleaseRepository task

mavencentral.gradle
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+}
+
+signing {
+    def signingKeyId = findProperty("signingKeyId")
+    def signingKey = findProperty("signingKey")
+    def signingPassword = findProperty("signingPassword")
+    useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
+    sign publishing.publications
```
gmerinojimenez, line 139: this is a workaround to be able to sign without exposing the key in any place. The key is stored in an [ascii armored format](https://www.techopedia.com/definition/23150/ascii-armor) in a secret in Vault and propagated here in an environment variable.

mavencentral.gradle
```diff
@@ -0,0 +1,140 @@
...
+                            dependencyNode.appendNode('artifactId', it.name)
+                            dependencyNode.appendNode('version', it.version)
+                        }
+                    }
+
+                    //Add a dependency with the library
+                    def dependencyNode = dependenciesNode.appendNode('dependency')
+                    dependencyNode.appendNode('groupId', PUBLISH_GROUP_ID)
+                    dependencyNode.appendNode('artifactId', PUBLISH_ARTIFACT_ID)
+                    dependencyNode.appendNode('version', PUBLISH_VERSION)
```
gmerinojimenez, line 114: otherwise we will add a dependency with com.telefonica.mistica:library, which is wrong

mbarrben (approved):
Great job!

dpastor (approved):
Nice work @gmerinojimenez

PR #88, opened 2021-06-18
ANDROID-9559 Add boxed inverse rows

mbarrben:
* [Add support for Boxed Inverse rows](https://jira.tid.es/browse/ANDROID-9559).
* Zeplin: https://zpl.io/bLe6AmQ

* Deprecate `listRowIsBoxed` attribute and `setBoxed` function. Deprecating it instead of removing it allows us to be backwards compatible and we won't break apps upgrading to this version of the library.
* Create new `ListRowView` property `backgroundType` that can be normal, boxed or boxedInverse.
* When background type is boxedInverse, we need to change TextView color to be inverse as well.
* Add a new `RecyclerView` to the list fragment in the catalog app to showcase boxed inverse rows.
* Extra ball: wrap all `RecyclerViews` in the list fragment inside a `NestedScrollView` so there's only one scroll. And fix the section titles.

### ☑️ Checks
- [X] I updated the [lists readme](https://github.com/Telefonica/mistica-android/tree/apps/ANDROID-9559_boxed_inverse_rows/library/src/main/java/com/telefonica/mistica/list).
- [X] Tested with dark mode.

- [X] 🖼️ Screenshots/Videos

Light | Dark
--- | ---
boxed_inverse_light | boxed_inverse_dark


catalog/src/main/res/layout/screen_fragment_lists_catalog.xml
```diff
@@ -1,40 +1,73 @@
...
-        android:id="@+id/boxed_list"
-        android:layout_width="match_parent"
-        android:layout_height="0dp"
-        android:layout_marginTop="8dp"
-        android:layout_weight="0.5"
-        app:listLayoutType="boxed" />
-
-</LinearLayout>
\ No newline at end of file
+<androidx.core.widget.NestedScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
mbarrben, line 2: Wrap all the recycler views inside a nested scroll view so it's easier to see all the lists. You can check the effect on the [readme gifs](https://github.com/Telefonica/mistica-android/tree/apps/ANDROID-9559_boxed_inverse_rows/library/src/main/java/com/telefonica/mistica/list)

catalog/src/main/res/layout/screen_fragment_lists_catalog.xml
```diff
@@ -1,40 +1,73 @@
...
-    android:layout_height="match_parent"
-    android:orientation="vertical">
-
-    <com.telefonica.mistica.section.SectionTitleView
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_gravity="start"
-        android:layout_marginStart="16dp"
-        android:layout_marginTop="16dp"
-        android:text="Full Width List" />
```
mbarrben, line 14: `android:text` doesn't work on section titles. Replaced with `app:sectionTitle`

catalog/src/main/res/layout/screen_fragment_lists_catalog.xml
```diff
@@ -1,40 +1,73 @@
...
+
+		<com.telefonica.mistica.list.MisticaRecyclerView
+				android:nestedScrollingEnabled="false"
+				android:id="@+id/boxed_inverse_list"
+				android:layout_width="match_parent"
+				android:layout_height="wrap_content"
+				android:layout_marginTop="8dp"
+				android:layout_weight="0.5"
+				app:listLayoutType="boxed"
+				/>
```
mbarrben, line 70: New list added

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -140,7 +159,18 @@ class ListRowView @JvmOverloads constructor(
...
+                BackgroundType.TYPE_BOXED
+            } else {
+                BackgroundType.TYPE_NORMAL
+            }
+            setBackgroundType(
+                styledAttrs.getInt(
+                    R.styleable.ListRowView_listRowBackgroundType,
+                    backgroundTypeDefaultValue
+                )
+            )
```
mbarrben, line 173: To be able to keep using the deprecated `listRowIsBoxed` attribute, we just use it to get the default value of the new `listRowBackgroundType` attribute. This way `listRowIsBoxed` works only if `listRowBackgroundType` is not being used.

library/src/main/res/drawable/boxed_inverse_list_row_background.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ripple xmlns:android="http://schemas.android.com/apk/res/android"
+    android:color="?attr/colorBackgroundAlternative">
+
+    <item android:id="@android:id/mask">
+        <shape>
+            <!-- Color is irrelevant for mask -->
+            <solid android:color="#000000" />
```
JorgeMucientes, line 8: Can't be removed if it doesn't have any effect?

JorgeMucientes:
Awesome job 🥇  

Although it feels kind of weird the design they have provided for inverse boxed style without any visible bounds.

library/src/main/res/drawable/boxed_inverse_list_row_background.xml
mbarrben: Maybe not the best comment: the thing is that you need to define a solid color for the mask to work, but it doesn't matter what color you use. I can change the comment to something like:
```xml
<!-- A color needs to be defined so the mask works, but it doesn't matter which color is used -->
```
WDYT?

JorgeMucientes: ah! ok ok. Yes that comment seem better

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -65,6 +65,11 @@ import com.telefonica.mistica.util.convertDpToPx
         attribute = "listRowIsBoxed",
         method = "setBoxed"
     ),
+    BindingMethod(
+        type = ListRowView::class,
+        attribute = "backgroundType",
+        method = "setBackgroundType"
+    ),
```
dpastor, line 72: Another option to avoid deprecating "isBoxed" could be adding a new boolean attr "hasInverseBackground".
But that would imply supporting inverse for non boxed rows, and anyway enum escalates better for future configurations.

dpastor (approved):
Good Job!

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
mbarrben: Yep, we thought about that case of non boxed + inverse, but there are no use cases right now. And if we need it in the future we can implement it with the enum as well.

mbarrben:
@JorgeMucientes let me know if you need anything else before approving this PR 😉

JorgeMucientes (approved):

JorgeMucientes:
> @JorgeMucientes let me know if you need anything else before approving this PR 😉

Sorry! no, I though I had approved it already.

PR #89, opened 2021-06-21
ANDROID-9497 Fixing promote action

gmerinojimenez:
[ANDROID-9497](https://jira.tid.es/browse/ANDROID-9497) Promote action was not working on remote (it did in local :( ), but in remote the sed command is not working
Just hardcoding the staging id, as it is not seems to be dangerous and it doesn't seems to change (I couldn't find documentation about it)

### ☑️ Checks

- [x] [Action result](https://github.com/Telefonica/mistica-android/actions/runs/956105758)


JorgeMucientes (approved):

JorgeMucientes:
No ticket associated?

gmerinojimenez:
> No ticket associated?

Not sure,

> No ticket associated?

added :)

build.gradle
```diff
@@ -17,10 +17,10 @@ buildscript {
 }
 
 apply plugin: 'io.codearte.nexus-staging'
+
 nexusStaging {
     packageGroup = "com.telefonica"
-    stagingProfileId = "com.telefonica"
-    stagingProfileId = System.getProperty("STAGING_PROFILE_ID")
+    stagingProfileId = "f7fe7699e57a"
```
manolovn, line 23: is safe to put this value here?

manolovn (approved):

build.gradle
gmerinojimenez: yes, there is no way to do anything without the authentication, and with it, this can be retrieved easily.

PR #91, opened 2021-07-02
ANDROID-9393 New empty state screen and card components

dpastor:

Implement empty state screen and card components in Mistica

Ticket: [ANDROID-9393](https://jira.tid.es/browse/ANDROID-9393)
Zeplin: [Empty States](https://zpl.io/b6Kdnoq)

* Implement screen and card components as independent custom views, with databinding support
* Create two new catalog entries for each component, following same conventions for existing components.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Update documentation

- [x] 🖼️ Screenshots/Videos

 
 

- [x] [Hockey](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-design-review/distribution_groups/public)

descarga


yceballost (changes requested):
I can't find it in code haha but, is description using textSecondary? because I think is applying textPrimary for description and this is wrong. (in empty state screen and empty state card)

The rest is perfect!

Extraball: was really nice to have an icon instead of image asset in Icon state on the example ^.^

library/src/main/res/layout/empty_state_card_view.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+		xmlns:tools="http://schemas.android.com/tools"
+		tools:parentTag="FrameLayout"
+		>
+
+	<LinearLayout
```
mbarrben, line 7: This `FrameLayout` has only one child, the `LinearLayout` right below. Is it possible to make the custom view a ´LinearLayout´ instead so we simplify the hierarchy a bit? Something like:

```xml
<merge xmlns:android="http://schemas.android.com/apk/res/android"
		xmlns:tools="http://schemas.android.com/tools"
		tools:parentTag="LinearLayout"
                android:layout_width="match_parent"
		android:layout_height="wrap_content"
		android:background="@drawable/empty_state_card_view_background"
		android:focusable="true"
		android:orientation="vertical"
		android:paddingTop="24dp"
		android:paddingBottom="24dp"
		>
    ...
```

library/src/main/res/layout/empty_state_card_view.xml
```diff
@@ -0,0 +1,92 @@
...
+
+		<LinearLayout
+				android:id="@+id/buttons_container"
+				android:layout_width="wrap_content"
+				android:layout_height="wrap_content"
+				android:gravity="start"
+				android:orientation="horizontal"
+				android:layout_marginTop="16dp"
+				android:layout_marginEnd="16dp"
+				android:layout_marginStart="0dp"
```
mbarrben, line 58: It's always weird to have asymmetry in the margins/paddings. In this case I guess it's because buttons inside may have different start margin. WDYT of configuring the start margin adding an extra 8dp to the link like:
```xml
<LinearLayout
    android:layout_marginEnd="16dp"
    android:layout_marginStart="16dp">

    <Button android:id="primary" />
    <Button android:id="secondary" />
    <Button
        android:id="link"
        android:layout_marginStart="8dp
    />
</LinearLayouta
```

library/src/main/res/layout/empty_state_screen_view.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+		xmlns:app="http://schemas.android.com/apk/res-auto"
+		xmlns:tools="http://schemas.android.com/tools"
+		tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout"
```
mbarrben, line 5: Can't this be just a vertical LinearLayout?

library/src/main/res/layout/empty_state_card_view.xml
dpastor: Sure, it can be done. I just used the FrameLayout in order to not duplicate LinearLayout configuration both programatically and in the layout, but you are right that at the end, hierarchy simplification weights more than this. Will upload change in a new commit :)

library/src/main/res/layout/empty_state_screen_view.xml
dpastor: Yes, like in card view, I used it because it does not require to configure orientation both programatically and in the layout, but it makes sense. Will change it in a new commit :)

library/src/main/res/layout/empty_state_card_view.xml
dpastor: This is the problematic scenario:



Link should be placed at 8dp, but from the card start, and 24dp when it is placed after any other button, while other buttons are always at 16dp.

dpastor:
> I can't find it in code haha but, is description using textSecondary? because I think is applying textPrimary for description and this is wrong. (in empty state screen and empty state card)
> 
> The rest is perfect!
> 
> Extraball: was really nice to have an icon instead of image asset in Icon state on the example ^.^

Totally correct @yceballost. I just missed that. I will upload it on a new commit. Thanks a lot.

library/src/main/res/layout/empty_state_card_view.xml
mbarrben: AFAIK you don't need to duplicate the layout configuration in Kotlin code, the merge tag does that for you.

library/src/main/res/layout/empty_state_card_view.xml
mbarrben: Oh! Understood! They key part is "and 24dp when it is placed after any other button". Thanks for the explanation!

mbarrben (approved):
good job!

manolovn:
nice job man!

manolovn (approved):

library/src/main/res/layout/empty_state_card_view.xml
dpastor: As talked by chat, attributes in merge tag are not applied to the custom view. In any case i will duplicate configuration using `tools` ns just to see the layout file correctly.

dpastor: https://github.com/Telefonica/mistica-android/pull/91/commits/88ef11634ccd7f4ae8dcc244f4465f9f4ecc86b7

library/src/main/res/layout/empty_state_screen_view.xml
dpastor: Fixed in https://github.com/Telefonica/mistica-android/pull/91/commits/88ef11634ccd7f4ae8dcc244f4465f9f4ecc86b7

yceballost (approved):
good job!!

dpastor:
Just re-requesting review for the latest 2 commits:
https://github.com/Telefonica/mistica-android/pull/91/commits/da2532b3baa82194d63d935ea8420431df7b55ba -> Fixes from detailed review with @yceballost. Just details found after checking component behaviour. Also makes actions/subtitle optional, which were incorrectly mandatory.
https://github.com/Telefonica/mistica-android/pull/91/commits/fe062aa88083b7efa33976a3584f839a1ad11bc9 -> Documentation

manolovn (approved):
nice! thx!

mbarrben (approved):

PR #93, opened 2021-07-08
ANDROID-9654 Fix list row icon margins

mbarrben:
Related to [this bug in the Novum app](https://jira.tid.es/browse/ANDROID-9654)

Icons in `ListRowView` had the wrong top margin on some cases. According to the [design](https://zpl.io/bzrKE63), there are two cases for the icon alignment:

1. The icon should be centered vertically in the row when only the title is visible, regardless of the type of the icon and the number of lines of the title.
2. The icon should have a 20dp margin from the top of the row when there are some other text visible besides the title, also regardless of the icon type.

in our implementation we were differentiating the second case. We were setting a 20dp top margin when the icon is large, and a 24dp top margin when the icon is small.

* Always set a 4dp top margin (that will be added to the 16dp parent padding to have a total of 20dp) when only the title is visible, regardless of the type of the icon.
* Add a couple of new samples to the catalog that showcase this scenario.
* Extra ball: Autoformat the list row view layout file.

### ☑️ Checks
- [X] No need to update docs, just a bug fix
- [X] No need to test with dark mode, only margins change.

- [X] 🖼️ Screenshots/Videos

#### Catalog app
Before | After
--- | ---
lists_before | lists_after

#### Novum app
Before | After
--- | ---
aura_before | aura_after


yamal-alm (approved):
nice catch!

mbarrben:
> nice catch!

Kudos to @cesarlopezfab

dpastor (approved):
Nice. Good Job!

PR #94, opened 2021-07-16
ANDROID-9693 Fix ListRowView text color on API < 23

mbarrben:
Fix this bug https://jira.tid.es/browse/ANDROID-9693 where `ListRowView` text have a wrong color on devices with API < 23.

There are actually 2 bugs here:
* For API < 23 the color is retrieved without taking the theme into account.
* We're applying primary color to subtitle and description, and it should be secondary [according to the docs](https://zpl.io/25LQ8ne).

`ContextCompat.getColor` ignores theme for API < 23, also for `ColorStateList` (selectors). So we need to find a different way to set the `ColorStateList` colors into the `TextViews`.
* Instead of retrieving the `ColorStateList` from resources, we create a new one for each case (thus duplicating those colors, now we have them defined in xml resources and Kotlin code. I'm open to better solutions).
* Apply primary or primary inverse colors to the title.
* Apply secondary or secondary inverse colors to the subtitle and the description.
* Extra ball: get rid of the all the magic numbers in `ListsCatalogFragment` by having a list of row configurations that we apply in the list adapter.
* Extra ball: Added new check to the PR template to test with API 21.

### ☑️ Checks
- [X] [Updated list docs](https://github.com/Telefonica/mistica-android/tree/apps/ANDROID-9693_listrow_text_color/library/src/main/java/com/telefonica/mistica/list)
- [X] Tested with dark mode.
- [X] Tested with API 21.

- [X] 🖼️ Screenshots/Videos

Light Before | Dark Before | API 21 Before
--- | --- | ---
android9693_before_light | android9693_before_dark | android9693_lists_before_api21

Light After | Dark After | API 21 After
--- | --- | ---
android9693_lists_after_light | android9693_lists_after_dark | android9693_lists_after_api21

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,6 +10,7 @@ _Provide a description of the implementation. A list of steps would be ideal._
 ### ☑️ Checks
 - [ ] I updated the documentation (readmes, wikis, etc). If no need to update documentation explain why.
 - [ ] Tested with dark mode.
+- [ ] Tested with API 21.
 
 ### :test_tube: How can I test this?
 _If it cannot be tested explain why._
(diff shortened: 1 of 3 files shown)


manolovn (approved):
thx!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,6 +10,7 @@ _Provide a description of the implementation. A list of steps would be ideal._
 ### ☑️ Checks
 - [ ] I updated the documentation (readmes, wikis, etc). If no need to update documentation explain why.
 - [ ] Tested with dark mode.
+- [ ] Tested with API 21.
```
gmerinojimenez, line 13: Good idea

gmerinojimenez (approved):
Thank you!!

PR #95, opened 2021-07-19
ANDROID-9679 Fix translations

mbarrben:
Fix https://jira.tid.es/browse/ANDROID-9679, where a copy is not properly translated to pt-BR

* Just fix the translations

### ☑️ Checks
- No need to update docs
- No need to test with dark mode or API 21

- [X] 🖼️ Screenshots/Videos

Before | After
--- | ---
Screenshot_2021-07-09-13-32-16-989_br com vivo enterprise | android9679_fix


yamal-alm (approved):
thanks 👍

gmerinojimenez (approved):

PR #99, opened 2021-09-09
ANDROID-9847 Fix colors in dialogs

gmerinojimenez:
Color primary was set to a color background (that was almost black). I'm setting the color primary in dark mode themes to the colorBrand.

### ☑️ Checks
- [x] Tested with dark mode.
- [ ] Tested with API 21.

Before:
image

After:
image

### DIFF
--- a/library/src/main/res/values/themes.xml
+++ b/library/src/main/res/values/themes.xml
@@ -5,7 +5,7 @@
 
     <style name="MisticaTheme_Base" parent="Theme.MaterialComponents.DayNight.NoActionBar">
 
-        <item name="colorPrimary">?colorNavigationBarBackground</item>
+        <item name="colorPrimary">?colorBrand</item>
         <item name="colorPrimaryDark">?colorBrandDark</item>
         <item name="android:windowBackground">?colorBackground</item>
         <item name="android:textColorPrimary">@color/text_primary_selector</item>


library/src/main/res/values/themes.xml
```diff
@@ -5,7 +5,7 @@
 
     <style name="MisticaTheme_Base" parent="Theme.MaterialComponents.DayNight.NoActionBar">
 
-        <item name="colorPrimary">?colorNavigationBarBackground</item>
+        <item name="colorPrimary">?colorBrand</item>
```
yamal-alm, line 8: How wasn't it affecting other parts of the app? 🤔  Maybe this change fixes other bugs that hasn't been reported yet..

yamal-alm (approved):
good catch!

library/src/main/res/values/themes.xml
dpastor: Looks like it was an issue, but this specific commit is confusing, as it clearly sets nav bar background color as the colorPrimary:
https://github.com/Telefonica/mistica-android/commit/e5bc61c1dc1af82e4b9a22f1bc61a9d1571a63f2

It's weird that no other things were affected, as @yamal-alm says.

It should have effect over dark mode colors, as the others are the same.

dpastor:
Adding @yceballost to the review.

library/src/main/res/values/themes.xml
gmerinojimenez: I want to check it in Android-messenger anyway, before integrating this change

dpastor (approved):

library/src/main/res/values/themes.xml
yceballost: maybe is better textLink color?

yceballost (approved):

gmerinojimenez:
I'm closing this review without integrating it, we have several places that are using the colorPrimary as colorNavigationBarBackground, like for example the action bar. I'm fixing the bug in the dialogs theme in another PR

PR #100, opened 2021-09-13
ANDROID-9856 Make mistica drawable loads to use AppCompatResources

dpastor:
Programatic drawable load should be done with `AppCompatResources`, in order to support new vector drawables attributes on old android api versions. This avoids possible crashes on older api versions.
For example, Empty states illustrations loads are crashing without this.

From [link](https://stackoverflow.com/a/43005362):
```
#1 Do you have to support API 4, 5, or 6?
    Yes: No choice but to use ResourcesCompat or ContextCompat.
    No: Keep going to #2.
#2 Do you absolutely need to supply a custom Theme?
    Yes: No choice but to use ResourcesCompat
    No: Use AppCompatResources
```

* Just change all programatic drawable loads to use `AppCompatResources`

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

- [x] Just checking image load behaves in same way


gmerinojimenez (approved):
Nice job!

PR #101, opened 2021-09-21
ANDROID-9698 Removing admin check in workflow_dispatch

gmerinojimenez:
Workflow-dispatch actions are only usable with write permissions. That should be enough for us.


dpastor (approved):
Nice! Good catch

yamal-alm (approved):

PR #102, opened 2021-09-27
ANDROID-9906 Update library module name

jeslat:
The goal of this change is to fix a problem when two libraries with the same module name are included in the same project. The file `META-INF/library_release.kotlin_module` has the same name in both cases and it throws an error.

* Just specifying a name to the library module `com.telefonica.mistica` in this case to avoid duplicates.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

imagen

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -19,6 +19,10 @@ android {
     defaultConfig {
         vectorDrawables.useSupportLibrary = true
     }
+
+    kotlinOptions {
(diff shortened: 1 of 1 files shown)


yamal-alm (approved):
thanks!

gmerinojimenez (approved):

PR #103, opened 2021-10-01
ANDROID-9928 [compose] Add support for Compose

gmerinojimenez:
Adding basic support for compose **Work in progress review**

Several things in this PR

- Adding support for compose in build.gradle
- I've created two new modules, library-compose and catalog-compose
- I've created a new launcher activity, so when installing the app you can see the two catalogs
- I've added a **very** basic catalog activity as a placeholder to start adding the migrated components
- As a PoC, I've wrapped the buttons (styles are not supported yet)
- Also created a MisticaTheme that allows us to use our material themes.
- Updated to Android 12

Pending things:
- Modify the github actions to export the two new artifacts, the library-compose and the catalog-compose to mavenCentral.
- Finish wrapping the buttons (maybe it's worth to analyze a bit if they can be rewritten in Compose, seems easy enough at least to be considered.
- ⚠️  I've seen that attributes in the drawables are not working well. We need to investigate this.  ⚠️

https://user-images.githubusercontent.com/4595241/135639819-5aebd5e9-eeac-40c0-b1f9-1553b73930ed.mp4


app/build.gradle
```diff
@@ -1,17 +1,21 @@
...
 
     defaultConfig {
         applicationId "com.telefonica.mistica.app"
         minSdkVersion 21
-        targetSdkVersion 29
+        targetSdkVersion 31
         versionCode versionCodeDate()
         versionName "1.0.0"
+
+        buildConfigField "String", "APPCENTER_KEY", "\"" + System.env.APPCENTER_KEY + "\""
```
gmerinojimenez, line 18: Moved to app module

app/src/main/AndroidManifest.xml
```diff
@@ -17,13 +17,27 @@
...
+                android:name="com.telefonica.mistica.compose.catalog.ui.CatalogMainActivity"
+                android:exported="true"
+                android:label="@string/compose_activity"
+                android:taskAffinity="com.telefonica.mistica.compose.catalog"
+                tools:node="replace">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+        </activity>
```
gmerinojimenez, line 39: The second catalog activity exposed in app

app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,4 @@
 <resources>
     <string name="app_name">Mística</string>
+    <string name="compose_activity">M.Compose</string>
```
gmerinojimenez, line 3: We need to find a better name :S

catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/CatalogMainActivity.kt
```diff
@@ -0,0 +1,142 @@
...
+        Icon(
+            painter = painterResource(id = R.drawable.icn_arrow),
+            contentDescription = null,
+            modifier = Modifier.align(Alignment.CenterEnd)
+        )
+    }
+}
+
+@Composable
+fun CatalogNavHost(navController: NavHostController, modifier: Modifier = Modifier) {
```
gmerinojimenez, line 91: this is the navigation definition. Only two screens right now

catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Buttons.kt
```diff
@@ -0,0 +1,9 @@
+package com.telefonica.mistica.compose.catalog.ui.components
+
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+
+@Composable
+fun Buttons() {
+    Text("Buttons!")
+}
```
gmerinojimenez, line 9: WIP 😅

catalog/build.gradle
```diff
@@ -1,17 +1,17 @@
...
+    id 'com.android.library'
+    id 'org.jetbrains.kotlin.android'
+    id 'kotlin-android-extensions'
+    id 'kotlin-kapt'
+    id 'maven-publish'
+}
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 31
```
gmerinojimenez, line 10: Android 12, mostly known as the Android 13-Torrijas predecessor

library-compose/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -0,0 +1,28 @@
...
+        factory = { context ->
+            Button(context).apply {
+                text = buttonText
+                icon = buttonIcon
+                setOnClickListener { onClickListener() }
+            }
+        },
+        update = { view -> view }
+    )
+}
```
gmerinojimenez, line 28: Also WIP, we need to overload the constructor to support custom styling

app/src/main/AndroidManifest.xml
```diff
@@ -17,13 +17,27 @@
 
         <activity
             android:name="com.telefonica.mistica.catalog.ui.CatalogMainActivity"
+            android:exported="true"
+            android:taskAffinity="com.telefonica.mistica.catalog"
```
gmerinojimenez, line 21: this is the trick to support both apps in the launcher, having two different affinities allows us to show them in different tasks in the launcher, like if they were two different apps

catalog-compose/src/main/res/drawable/ic_buttons.xml
```diff
@@ -0,0 +1,29 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
yamal-alm, line 1: are these assets duplicated from catalog? WDYT about a catalog-commons? maybe by now it doesn'tt make sense btw..

yamal-alm (approved):
Thanks Guille, it looks good! Let me complete a few more codelabs and I'll help you with the support for compose 😄

catalog-compose/src/main/res/drawable/ic_buttons.xml
gmerinojimenez: the catalog is not supposed to be in any release application, so I think it's not a problem to have them duplicated (better than have to support an additional module (that increases build time))

Anyway I don't have a strong opinion

yamal-alm: fair enough 👍

manolovn (approved):
Greeeeeeat!! thanks a lot!

PR #117, opened 2021-11-03
ANDROID-10011 Set max lines to CardView title and description

pmartinbTEF:
Add max lines parameter for title and description in CardView

Added cardTitleMaxLines and cardDescriptionMaxLines atributes

### ☑️ Checks
- [ ] I updated the corresponding readmes
- [ ] Tested in Smart WiFi with this tutorial: https://github.com/Telefonica/android-messenger/wiki/Testing-libraries

Before
drawing

After
drawing


library/src/main/java/com/telefonica/mistica/card/CardView.kt
```diff
@@ -32,11 +34,21 @@ import com.telefonica.mistica.util.getThemeColor
         attribute = "cardTitle",
         method = "setTitle"
     ),
+    BindingMethod(
+        type = MediaCardView::class,
```
jeslat, line 38: I think the type should be `CardView::class`, shouldn't be?

library/src/main/java/com/telefonica/mistica/card/CardView.kt
```diff
@@ -32,11 +34,21 @@ import com.telefonica.mistica.util.getThemeColor
...
+        attribute = "cardTitleMaxLines",
+        method = "setTitleMaxLines"
+    ),
     BindingMethod(
         type = CardView::class,
         attribute = "cardDescription",
         method = "setDescription"
     ),
+    BindingMethod(
+        type = MediaCardView::class,
```
jeslat, line 48: Same here

yamal-alm (approved):
nice!

gmerinojimenez (approved):
Nice work!

library/src/main/java/com/telefonica/mistica/card/CardView.kt
pmartinbTEF: Done

library/src/main/java/com/telefonica/mistica/card/CardView.kt
pmartinbTEF: Done

jeslat (approved):
good job!

PR #120, opened 2021-11-12
Set default text color for compose themes

dpastor:
Just set default text color on compose themes, so brand and dark mode config changes modifies it.

* Just configure base text style with a default color

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

- [x] 🖼️ Screenshots/Videos
Captura de pantalla 2021-11-12 a las 13 10 21

### DIFF
--- a/library-compose/src/main/java/com/telefonica/mistica/compose/theme/text/MisticaTypography.kt
+++ b/library-compose/src/main/java/com/telefonica/mistica/compose/theme/text/MisticaTypography.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.runtime.structuralEqualityPolicy
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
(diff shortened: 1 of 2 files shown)


gmerinojimenez (approved):
Nice job!

yamal-alm (approved):
nice 👏

library-compose/src/main/java/com/telefonica/mistica/compose/theme/text/MisticaTypography.kt
```diff
@@ -5,13 +5,15 @@ import androidx.compose.runtime.mutableStateOf
...
 import androidx.compose.runtime.structuralEqualityPolicy
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.sp
 
 class MisticaTypography(
     private var fontFamily: FontFamily = FontFamily.SansSerif,
+    private var defaultTextColor: Color = Color.Unspecified,
```
yamal-alm, line 16: Just curiosity: if you let the method use this default Undefined value, how are the texts displayed?

dpastor: It is also the default value of text style constructor -->
```kotlin
@Immutable
class TextStyle(
    val color: Color = Color.Unspecified,
    ...
) {
```
It looks like some kind of NullObject for colors, in case nothing is defined it is rendered with black color as I see.

PR #126, opened 2021-11-16
ANDROID-10054 - Deleting app tag

gmerinojimenez:
We're exporting `supportRtls = false` attribute in Mística. This ticket fixes the errors that doesn't enable us to set it to true and then removes it from the manifest.

 I've fixed the lint error (only one) so we can support Rtls (according to Lint).

Then I've excluded the supportRtls tag from the library manifest by ignoring the lint rule that checks it.

The rule was forcing me to put it in the manifest:
```
/Users/gmerino/Workspace/tuenti/mistica-android/library/src/main/AndroidManifest.xml: Error: The project references RTL attributes, but does not explicitly enable or disable RTL support with android:supportsRtl in the manifest [RtlEnabled]
```

- Having blue ball should be enough
- Tested lists manually

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -11,7 +11,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
-        android:supportsRtl="false"
+        android:supportsRtl="true"
         android:name=".CatalogApplication"
         android:theme="@style/MisticaTheme.Movistar">
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -11,7 +11,6 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
-        android:supportsRtl="true"
         android:name=".CatalogApplication"
         android:theme="@style/MisticaTheme.Movistar">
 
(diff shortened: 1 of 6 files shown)


yamal-alm (approved):
<div dir="rtl">
Great, Thanks!
</div>

app/src/main/AndroidManifest.xml
```diff
@@ -11,7 +11,6 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
-        android:supportsRtl="false"
```
nimeacuerdo, line 14: My only doubt here was about the effect of this change given this attribute already has a default value of `false` according to [docs](https://developer.android.com/guide/topics/manifest/application-element#supportsrtl):

> Declares whether your application is willing to support right-to-left (RTL) layouts.
> If set to true and targetSdkVersion is set to 17 or higher, various RTL APIs will be activated and used by the system so your app can display RTL layouts. If set to false or if targetSdkVersion is set to 16 or lower, the RTL APIs will be ignored or will have no effect and your app will behave the same regardless of the layout direction associated to the user's Locale choice (your layouts will always be left-to-right).
> 
> The default value of this attribute is false.
> 
> This attribute was added in API level 17.

gmerinojimenez: but in this way there is no <application> tag node in our library. So there is no default value for the library's generated aar. My point is that it should be specified (or use the default value) in the application module that uses mística.

image

I deleted it from the app AndroidManifest, but I can restore it, maybe it's more clear in this way (in any case the app module is never exported, only the catalog and the library)

dpastor (approved):

PR #129, opened 2021-11-26
ANDROID-10116 - Remove library dependency from the pom.xml and adding Mistica instead

gmerinojimenez:
If you check https://search.maven.org/artifact/com.telefonica/mistica-compose/3.2.0/aar
you'll see 
image
We should use "mistica" instead


manolovn (approved):
cool! :rocket:

PR #130, opened 2021-11-29
Modifier to buttons

gmerinojimenez:
Adding a modifier parameter in the Button composable

Adding the modifier and propagating it to the AndroidView
Also some minor fixes in the UI and some imports that were wrong
### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

image


yamal-alm (approved):
Thanks ❤️

manolovn (approved):
nice! thanks!!

PR #132, opened 2021-11-30
ANDROID-10046 Error reference in Feedbacks

dpastor:
Add the possibility of including an error reference text in Error Feedbacks.
Check ticket for details -> ANDROID-10046

As talked with iOS, component will accept whole field text, instead of just the error reference, so, this way mistica does not need to hold any translation/default token.

* Include a new field in the feedback component that allows specifying a string for this
* Adapt compose function with this new parameter

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

- [x] 🖼️ Screenshots/Videos

  

- [x] [Hockey](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-core-alpha-1/distribution_groups/public)

  Captura de pantalla 2021-11-30 a las 13 05 15


yceballost (approved):
lgtm! thanks!

yamal-alm (approved):
nice! good job 👏

PR #134, opened 2021-12-01
ANDROID-10148 PoC of how to restrict @Composable functions to only allow certain components

gmerinojimenez:
This is just kind of a PoC, I hadn't time to migrate other usages, but I think it can be a good approach to restrict that some parts of bigger components that are defined as composable to accept only the components we want to add.

Limit list row headline to only accept Tags

* Expose a class with the same name as the composable function: Tag
* That class will have a build method that creates the composable
* Restrict the ListRowItem parameter to only accept Tag (the class)

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

https://user-images.githubusercontent.com/4595241/144257978-1e33b37d-fd6e-4f48-b86e-ae2aa5e95e71.mp4

### DIFF
--- a/catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
+++ b/catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
@@ -32,7 +32,8 @@ const val TITLE = "Title"
 const val SUBTITLE = "Subtitle"
 const val DESCRIPTION = "Description"
 
-val samples = listOf(
+@Composable
+fun samples() = listOf(
     ListItem(
         title = TITLE,
     ),
@@ -130,14 +131,14 @@ val samples = listOf(
     ),
 
     ListItem(
-        headline = { Tag("PROMO", color = MisticaTheme.colors.promo) },
+        headline = Tag(content ="PROMO").withColor(MisticaTheme.colors.promo),
         title = TITLE,
         subtitle = SUBTITLE,
         action = { Chevron() },
         icon = { ListIcon() },
     ),
     ListItem(
-        headline = { Tag("PROMO", color = MisticaTheme.colors.promo) },
+        headline = Tag("PROMO").withColor(MisticaTheme.colors.promo),
         title = TITLE,
         subtitle = SUBTITLE,
         description = DESCRIPTION,
@@ -147,7 +148,7 @@ val samples = listOf(
         icon = { Avatar() },
     ),
     ListItem(
-        headline = { Tag("PROMO", color = MisticaTheme.colors.promo) },
+        headline = Tag("PROMO").withColor(MisticaTheme.colors.promo),
         title = TITLE,
         subtitle = SUBTITLE,
         description = DESCRIPTION,
@@ -157,7 +158,7 @@ val samples = listOf(
         icon = { Avatar() },
     ),
     ListItem(
-        headline = { Tag("PROMO", color = MisticaTheme.colors.promo) },
+        headline = Tag("PROMO").withColor(MisticaTheme.colors.promo),
         title = TITLE,
         subtitle = SUBTITLE,
         description = DESCRIPTION,
@@ -172,6 +173,7 @@ val samples = listOf(
 @ExperimentalMaterialApi
 @Composable
 fun Lists() {
+    val samples = samples()
     LazyColumn(
         modifier = Modifier
             .fillMaxSize(),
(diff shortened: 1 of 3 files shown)


catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
```diff
@@ -248,7 +250,7 @@ data class ListItem(
     val backgroundType: BackgroundType = BackgroundType.TYPE_NORMAL,
     val badge: String? = null,
     val isBadgeVisible: Boolean = false,
-    val headline: @Composable (() -> Unit)? = null,
+    val headline: Tag? = null,
```
gmerinojimenez, line 253: This is the objective of the PR

library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -47,7 +47,7 @@ fun ListRowItem(
     backgroundType: BackgroundType = BackgroundType.TYPE_NORMAL,
     badge: String? = null,
     isBadgeVisible: Boolean = false,
-    headline: @Composable (() -> Unit)? = null,
+    headline: Tag? = null,
```
gmerinojimenez, line 50: This is the objective of the PR

library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -118,7 +118,7 @@ fun ListRowItem(
                     .align(CenterVertically)
             ) {
                 headline?.let {
-                    it()
+                    it.build()
```
gmerinojimenez, line 121: and here we build the composable

library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -177,7 +177,7 @@ fun ListRowItemPreview() {
         val checkedState = remember { mutableStateOf(true) }
         Column {
             ListRowItem(
-                headline = { Tag("promo") },
+                headline = Tag("Promo"),
```
gmerinojimenez, line 180: as you can see the usage is almost identical

library-compose/src/main/java/com/telefonica/mistica/compose/tag/Tag.kt
```diff
@@ -39,6 +39,22 @@ fun Tag(
     }
 }
 
+class Tag constructor(
+    private val content: String,
```
gmerinojimenez, line 43: content is mandatory, the other 2 are optional.

I'm not adding them to the constructor since it will collide with the fun Tag signature

dpastor (approved):
Nice approach :)

manolovn (approved):

library-compose/src/main/java/com/telefonica/mistica/compose/tag/Tag.kt
```diff
@@ -39,6 +39,22 @@ fun Tag(
...
+) {
+    private var modifier: Modifier = Modifier
+    private var color: Color? = null
+
+    fun withModifier(modifier: Modifier): Tag = this.apply { this.modifier = modifier }
+
+    fun withColor(color: Color): Tag = this.apply { this.color = color }
+
+    @Composable
+    fun build() {
```
dagonco, line 53: The linter is gonna complain, this method should start with a capital letter.
How do we manage this? Suppressing the warning with the annotation?
Or maybe we add some rule _(that I don't know if it's possible)_ so that the linter ignores the `build()` methods.

dagonco (approved):
Fancy implementation. I'll follow the same pattern in compose tag branch!

library-compose/src/main/java/com/telefonica/mistica/compose/tag/Tag.kt
gmerinojimenez: good point, if we do this for more components maybe it's worth to check if build methods can be ignored

gmerinojimenez: Hmm. After double thinking it, I think I prefer to suppress it: this method (or these if we add more builders) are going to be internal to the library and I think it's better to suppress them internally without adding a rule that can end up skipping functions named build that should be uppercase 
Also, I've changed the visibility of the method to internal :)

PR #140, opened 2022-01-19
ANDROID-10252 [Compose] Fix ListRowItem vertical padding and title style

yamal-alm:
Fix ListRowItem vertical padding and title style (only in Compose version of the component)

- Row title from light to regular
- Space between rows should be 16dp, when each row sum 8dp
- Removed dividers from catalog in order to be aligned with non-composable catalog where there aren't dividers

| Before | After |
| ------ | ------ |
| Captura de pantalla 2022-01-19 a las 16 27 29 | Captura de pantalla 2022-01-19 a las 16 20 09 |


yceballost (approved):
nice!

manolovn (approved):
Thx!!! :rocket:

pmartinbTEF (approved):
Good job!

PR #141, opened 2022-01-20
ANDROID-10257 - Not uploading mappings (they doesn't exist in debug)

gmerinojimenez:
ANDROID-10257 The appcenter plugin has a bug that crashes the upload when attempting to upload the mapping file if it doesn't exist

In Mística we will never have mapping file. So I'm setting the upload mapping file flag to false

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -39,6 +39,7 @@ appcenter {
             appName = "mistica"
         }
     }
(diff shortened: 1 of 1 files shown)


yamal-alm (approved):
thanks

nimeacuerdo (approved):
🚀

PR #144, opened 2022-01-28
ANDROID-10301 Fix link button in cards

gmerinojimenez:
ANDROID-10301 The link button is not shown if the action is not present

Replacing and `&&` by an `||` and adjusting the padding

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

image

### DIFF
--- a/library-compose/src/main/java/com/telefonica/mistica/compose/card/Card.kt
+++ b/library-compose/src/main/java/com/telefonica/mistica/compose/card/Card.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.offset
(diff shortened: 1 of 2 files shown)


dagonco (approved):
Thanks @gmerinojimenez !

library-compose/src/main/java/com/telefonica/mistica/compose/card/Card.kt
```diff
@@ -54,7 +55,7 @@ fun Card(
 
 @Composable
 internal fun CardActions(primaryButton: Action?, linkButton: Action?) {
-    if (primaryButton != null && linkButton != null) {
+    if (primaryButton != null || linkButton != null) {
```
yamal-alm, line 58: does non-compose component have the same condition?

yamal-alm (approved):
thanks!

library-compose/src/main/java/com/telefonica/mistica/compose/card/Card.kt
gmerinojimenez: yes, the && was my fault

PR #145, opened 2022-02-02
ANDROID-10322 Allow DataCard icons to be remote images, resources or initials

gmerinojimenez:
Adding a builder to be able to create different types of icons

**Initials**
image

**Resource**
image

**Url**
image


library-compose/src/main/java/com/telefonica/mistica/compose/card/datacard/DataCard.kt
```diff
@@ -77,4 +64,97 @@ fun CardPreview() {
             linkButton = Action("Link") {}
         )
     }
-}
\ No newline at end of file
+}
+
+sealed class IconPainter {
```
yamal-alm, line 69: I would move this class and Composable to a separate file WDYT

library-compose/src/main/java/com/telefonica/mistica/compose/card/datacard/DataCard.kt
```diff
@@ -77,4 +64,97 @@ fun CardPreview() {
...
+                    contentDescription = null,
+                    modifier = Modifier.size(40.dp)
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun resourceIconPainter(@DrawableRes iconRes: Int) = IconPainter.ResourceIconPainter(iconRes)
```
yamal-alm, line 143: Why is `@Composable` annotation needed?

library-compose/src/main/java/com/telefonica/mistica/compose/theme/color/MisticaColors.kt
```diff
@@ -235,6 +237,10 @@ class MisticaColors(
         internal set
     var brandDark by mutableStateOf(brandDark, structuralEqualityPolicy())
         internal set
+    var initialsIconBackground by mutableStateOf(initialsIconBackground, structuralEqualityPolicy())
+        internal set
+    var initialsIconForeground by mutableStateOf(initialsIconForeground, structuralEqualityPolicy())
+        internal set
```
yamal-alm, line 243: as discussed offline, this colors should not be in Mistica. Instead we are declaring them as mandatory params in `TextIconPainter`

yamal-alm (approved):
thanks, looks nice 😄

library-compose/src/main/java/com/telefonica/mistica/compose/card/datacard/DataCard.kt
gmerinojimenez: oki, sure

library-compose/src/main/java/com/telefonica/mistica/compose/card/datacard/DataCard.kt
gmerinojimenez: it was needed for accessing the themed colors, Since that's removed I'm removing the annotation as well

library-compose/src/main/java/com/telefonica/mistica/compose/theme/color/MisticaColors.kt
gmerinojimenez: Agree, removed

yceballost:
Can you share a link or image to see it?

Make sense to attend this as a independent component? a circle + initials to be used in any place. If yes, we have to add it to design library and make the same for mistica-web!

gmerinojimenez:
> Can you share a link or image to see it?
> 
> Make sense to attend this as a independent component? a circle + initials to be used in any place. If yes, we have to add it to design library and make the same for mistica-web!

Sure, added :)

dagonco (approved):
Awesome! Thanks :D

PR #148, opened 2022-02-08
ANDROID-10345 Do not capitalize data card subtitle in compose

yamal-alm:
Non-composable DataCard don't force caps on its subtitle so we shouldn't be doing the same in the Compose version.

Remove call to `uppercase` over DataCard subtitle

- [x] 🖼️ Screenshots/Videos
## Before
Captura de pantalla 2022-02-08 a las 16 40 45
## After
Captura de pantalla 2022-02-08 a las 16 40 00
## Non Compose DataCard
Captura de pantalla 2022-02-08 a las 16 41 03


dagonco (approved):
Thanks! 😁

PR #152, opened 2022-02-23
ANDROID-10425 - Set min height in lists

gmerinojimenez:
There is no minimum height. Setting one.
https://app.zeplin.io/project/5d653c69f828bf7299c551c1/screen/5e9d50079630597dd4731f3d

Just setting the min height

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

https://user-images.githubusercontent.com/4595241/155322587-19061450-0ed9-4232-9cfc-01ba7adee9d4.mp4

### DIFF
--- a/library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
+++ b/library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
@@ -30,9 +30,9 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.telefonica.mistica.R
 import com.telefonica.mistica.compose.badge.Badge
-import com.telefonica.mistica.compose.tag.Tag
 import com.telefonica.mistica.compose.shape.Chevron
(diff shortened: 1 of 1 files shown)


library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -54,7 +54,7 @@ fun ListRowItem(
     val badgeVisible by remember { mutableStateOf(isBadgeVisible) }
 
     val boxModifier = when (backgroundType) {
-        BackgroundType.TYPE_NORMAL -> Modifier
+        BackgroundType.TYPE_NORMAL -> modifier
```
gmerinojimenez, line 57: otherwise a new modifier is created, instead of using the one provided as parameter

manolovn: nice

manolovn (approved):
thanks!

dagonco (approved):
Nice 👌🏻

PR #156, opened 2022-03-22
Small changes in DropDownInput and  ListRowItem (compose)

yamal-alm:
Small fixes in compose components `DropDownInput` and `ListRowItem`. Also update docs.

- `DropDownInput`: Make `currentItemIndex` nullable so the client doesnt' have to explicitly specify a negative index in case a preselected value is not wanted. Also moved to correct package
- `ListRowItem`: make `onClick` lambda nullable so the ripple effect can be avoided.
- Update docs about what components are migrated
- `TextInput`, `PhoneInput` and `EmailInput` now they expose `readOnly` and `enabled` params (with default values) just like the built in `TextField, I think they could be useful. Also I've added an optional `onClick` listener, there are some cases where it could be needed.

### ☑️ Checks
- [x] Tested with dark mode.


library-compose/src/main/java/com/telefonica/mistica/compose/input/DropDownInput.kt
```diff
@@ -12,7 +12,7 @@ import com.telefonica.mistica.input.DropDownInput
 fun DropDownInput(
     modifier: Modifier = Modifier,
     items: List<String>,
-    currentItemIndex: Int,
+    currentItemIndex: Int? = null,
```
yamal-alm, line 15: this null, as a default value, doesn't force the developer to explicitly set a negative index to not select any item. I could have set `-1` as the default value but IMO null is more meaningful WDYT

library-compose/src/main/java/com/telefonica/mistica/compose/input/EmailInput.kt
```diff
@@ -0,0 +1,34 @@
+package com.telefonica.mistica.compose.input
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.input.KeyboardType
+
+@Composable
+fun EmailInput(
```
yamal-alm, line 8: just moved to separated file

library-compose/src/main/java/com/telefonica/mistica/compose/input/TextInput.kt
```diff
@@ -1,191 +1,34 @@
...
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.focus.onFocusChanged
-import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.unit.dp
-import com.telefonica.mistica.compose.theme.MisticaTheme
+import androidx.compose.ui.text.input.KeyboardType
 
 @Composable
-internal fun TextInputImpl(
+fun TextInput(
```
yamal-alm, line 8: ignore this change, `TextInputImpl` was in a file named `TextInput`. I have moved it to a file named `TextInputImpl` and moved `TextInput` to this file with the same name

library-compose/src/main/java/com/telefonica/mistica/compose/input/TextInputImpl.kt
```diff
@@ -0,0 +1,191 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.onFocusChanged
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import com.telefonica.mistica.compose.theme.MisticaTheme
+
+@Composable
+internal fun TextInputImpl(
```
yamal-alm, line 33: not new, just moved, see my previous comment

library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -173,6 +173,13 @@ fun ListRowItem(
     }
 }
 
+private fun Modifier.makeClickableIfNeeded(onClick: (() -> Unit)?): Modifier =
+    if (onClick != null) {
+        clickable(onClick = onClick)
+    } else {
+        this
+    }
```
yamal-alm, line 181: if null `onClick` is passed, then the clickable modifier is not added and then ripple effect is avoided

library-compose/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -49,7 +49,7 @@ fun ListRowItem(
     isBadgeVisible: Boolean = false,
     headline: Tag? = null,
     trailing: @Composable (() -> Unit)? = null,
-    onClick: () -> Unit = {},
+    onClick: (() -> Unit)? = {},
```
yamal-alm, line 52: I've kept the same default value to respect retrocompatibility

library-compose/src/main/java/com/telefonica/mistica/compose/input/TextInput.kt
```diff
@@ -1,191 +1,40 @@
...
-    label: String,
+    helperText: String? = null,
     isError: Boolean = false,
-    trailingIcon: @Composable (() -> Unit)?,
-    keyboardOptions: FoundationKeyboardOptions,
+    errorText: String? = null,
+    trailingIcon: @Composable (() -> Unit)? = null,
+    isInverse: Boolean = false,
+    enabled: Boolean = true,
+    readOnly: Boolean = false,
```
yamal-alm, line 19: two new lines

library-compose/src/main/java/com/telefonica/mistica/compose/input/EmailInput.kt
```diff
@@ -0,0 +1,40 @@
...
+    onValueChange: (String) -> Unit,
+    label: String,
+    helperText: String? = null,
+    isError: Boolean = false,
+    errorText: String? = null,
+    trailingIcon: @Composable (() -> Unit)? = null,
+    isInverse: Boolean = false,
+    enabled: Boolean = true,
+    readOnly: Boolean = false,
+    onClick: (() -> Unit)? = null,
```
yamal-alm, line 20: three new lines

library-compose/src/main/java/com/telefonica/mistica/compose/input/TextInputImpl.kt
```diff
@@ -0,0 +1,203 @@
...
+    onClick: (() -> Unit)?,
+    keyboardOptions: FoundationKeyboardOptions,
+) {
+    val interactionSource = remember {
+        MutableInteractionSource()
+    }
+
+    if (onClick != null && interactionSource.collectIsPressedAsState().value) {
+        onClick()
+    }
```
yamal-alm, line 98: `interactionSource` is a `StateFlow` which emits events regarding interactions performed over the `TextField. Also, this `onClick` will be called just once

gmerinojimenez (approved):
Nice job!

PR #158, opened 2022-03-29
Wrap IndterminateStepperView into AndroidView composable

yamal-alm:
Migrate Stepper to compose, but for now only `IndeterminateStepperView` view since it is needed by another app.

- Wrap `IndeterminateStepperView` into `IndeterminateStepper`
- Add param `animate: Boolean = true` to `IndeterminateStepperView::setProgress` method because there are some places where animation needs to be explicitly disabled

### ☑️ Checks
- [x] Tested with dark mode.

- [x] [Mistica App download link](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public)
- [x] 🖼️ Screenshots/Videos

https://user-images.githubusercontent.com/47142570/160644415-5d879efd-09e4-45de-81d2-97ed609520ad.mov

### DIFF
--- a/catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Steppers.kt
+++ b/catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Steppers.kt
@@ -63,19 +63,21 @@ private fun IndeterminateStepperDemo(modifier: Modifier) {
             Button(
                 text = "Previous",
                 onClickListener = {
-                    if (progress - 10 >= 0) {
-                        progress -= 10
+                    if (progress - STEPPER_DEMO_PROGRESS_INTERVAL >= 0) {
+                        progress -= STEPPER_DEMO_PROGRESS_INTERVAL
                     }
                 }
             )
             Button(
                 text = "Next",
                 onClickListener = {
-                    if (progress + 10 <= 100) {
-                        progress += 10
+                    if (progress + STEPPER_DEMO_PROGRESS_INTERVAL <= 100) {
+                        progress += STEPPER_DEMO_PROGRESS_INTERVAL
                     }
                 }
             )
         }
     }
 }
+
+const val STEPPER_DEMO_PROGRESS_INTERVAL = 20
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/telefonica/mistica/stepper/DeterminateStepperView.kt
```diff
@@ -170,7 +170,7 @@ class DeterminateStepperView @JvmOverloads constructor(
         if (isDetermined) {
             setDeterminedStep(step)
         } else {
-            setIndeterminateStep(step)
+            setIndeterminateStep(step, animate = true)
```
yamal-alm, line 173: when the view is initialized, the progress shouldn't be animated because the view is expected to already be set to a specific progress

library/src/main/java/com/telefonica/mistica/stepper/IndeterminateStepperView.kt
```diff
@@ -40,18 +40,22 @@ class IndeterminateStepperView @JvmOverloads constructor(
             )
 
             val progress = styledAttrs.getInteger(R.styleable.IndeterminateStepperView_progress, 0)
-            setProgress(progress)
+            setProgress(progress, animate = false)
```
yamal-alm, line 43: when the view is created, the progress shouldn't be animated because the view is expected to already be set to a specific progress

gmerinojimenez (approved):
Good job!

AnaMontes11 (commented):
The progress works nicely but the animation is not fluid. @yamal-alm

catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Steppers.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.telefonica.mistica.compose.stepper.IndeterminateStepper
+
+@Composable
+fun Steppers() {
+    val scrollState = rememberScrollState()
+
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
```
pmartinbTEF, line 28: padding could be extracted to constant (same for the progress 10% step)

AnaMontes11 (approved):
Amazing! 🚀

PR #160, opened 2022-05-25
ANDROID-10695 Make Feedback onBackpressed param optional

yamal-alm:
`onBackPressed` param from `Feedback` composable component should be optional so the back event is not overridden when needed.

- Make `onBackPressed` nullable with a default `null` value and only explicitly handle back event when it is present.
- Also add `null` default values to second button text and its click listener

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

- [x] Tested manually that passing null as `onBackPressed` do not override the back event

### DIFF
--- a/library-compose/src/main/java/com/telefonica/mistica/compose/feedback/Feedback.kt
+++ b/library-compose/src/main/java/com/telefonica/mistica/compose/feedback/Feedback.kt
@@ -1,16 +1,12 @@
 package com.telefonica.mistica.compose.feedback
 
 import android.view.ViewGroup
-import androidx.activity.OnBackPressedCallback
-import androidx.activity.compose.LocalOnBackPressedDispatcherOwner
+import androidx.activity.compose.BackHandler
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
-import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.compose.ui.viewinterop.AndroidView
 import com.telefonica.mistica.feedback.screen.view.FeedbackScreenView
 
@@ -31,7 +27,7 @@ fun Feedback(
     var shouldPerformTheAnimation by remember { mutableStateOf(true) }
 
     onBackPressed?.let {
-        BackHandler(it)
+        BackHandler(onBack = onBackPressed)
     }
 
     AndroidView(factory = { context ->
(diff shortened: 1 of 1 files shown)


library-compose/src/main/java/com/telefonica/mistica/compose/feedback/Feedback.kt
```diff
@@ -21,36 +21,17 @@ fun Feedback(
...
-    val lifecycleOwner = LocalLifecycleOwner.current
-
-    DisposableEffect(lifecycleOwner, backDispatcher) {
-        backDispatcher.addCallback(lifecycleOwner, backCallback)
-        onDispose {
-            backCallback.remove()
-        }
+    onBackPressed?.let {
+        BackHandler(it)
     }
```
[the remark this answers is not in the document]
manolovn: can we use the already inlcuded in compose? https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#backhandler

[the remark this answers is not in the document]
yamal-alm: totally

library-compose/src/main/java/com/telefonica/mistica/compose/feedback/Feedback.kt
```diff
@@ -21,36 +17,17 @@ fun Feedback(
...
-    val backDispatcher = checkNotNull(LocalOnBackPressedDispatcherOwner.current) {
-        "No OnBackPressedDispatcherOwner was provided via LocalOnBackPressedDispatcherOwner"
-    }.onBackPressedDispatcher
-    val lifecycleOwner = LocalLifecycleOwner.current
-
-    DisposableEffect(lifecycleOwner, backDispatcher) {
-        backDispatcher.addCallback(lifecycleOwner, backCallback)
-        onDispose {
-            backCallback.remove()
-        }
```
yamal-alm, line 53: this code is actually the same implementation included in existing `androidx.activity.compose.BackHandler`

manolovn (approved):
thx! :rocket:

library-compose/src/main/java/com/telefonica/mistica/compose/feedback/Feedback.kt
```diff
@@ -1,79 +1,61 @@
...
-import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
-import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.compose.ui.Modifier
 import androidx.compose.ui.viewinterop.AndroidView
 import com.telefonica.mistica.feedback.screen.view.FeedbackScreenView
 
 @Composable
 fun Feedback(
+    modifier: Modifier = Modifier,
```
yamal-alm, line 16: I've added a `modifier` param so we can do things like adding a `testTag` modifier

manolovn: nice! thx!

gmerinojimenez (approved):
Thank you!!

PR #174, opened 2022-08-29
ANDROID-10946 Bottom sheet dialog

gmerinojimenez:
[ANDROID-10946](https://jira.tid.es/browse/ANDROID-10946)

See specs here: 
https://www.figma.com/file/nxToAYk7FGC86TS0mY61TU/%5BWIP%5D-Sheet?node-id=2%3A61

### ☑️ Checks
- [X] Tested with dark mode.
- [ ] Tested with API 21.

https://user-images.githubusercontent.com/4595241/187227051-12506aec-75a6-45c1-9380-68f628d23cfe.mp4


catalog/build.gradle
```diff
@@ -23,6 +23,9 @@ android {
     lintOptions {
         warningsAsErrors true
     }
+    viewBinding {
+        enabled = true
+    }
```
gmerinojimenez, line 28: added only in the catalog

catalog/src/main/res/layout/activity_main_catalog.xml
```diff
@@ -2,6 +2,7 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
+    tools:viewBindingIgnore="true"
```
gmerinojimenez, line 5: I've added this tag to the catalog layouts

library/src/main/java/com/telefonica/mistica/bottomsheet/BottomSheetModel.kt
```diff
@@ -0,0 +1,42 @@
...
+    data class RowWithCheckBox(
+        override val id: String,
+        val title: String,
+        val description: String? = null,
+        val asset: Asset? = null,
+        val selected: Boolean = false,
+    ) : RowWithCheckboxElement(
+        id = id,
+    )
+}
```
yamal-alm, line 36: does `RowWithCheckboxElement` really need to be a sealed class? isn't it too specific already? 
Also, I don't understand why we need a list of lists (a list of children), can we have instead just a list of elements? where by now there is only one type: RowWithCheckboxElement (inheriting from a RowElement sealed class e.g.)

library/src/main/java/com/telefonica/mistica/bottomsheet/BottomSheet.kt
```diff
@@ -0,0 +1,162 @@
...
+    ): BottomSheet = this.apply {
+        bottomSheetModel = bottomSheetModel.copy(header = Header(title, subtitle, description))
+    }
+
+    fun withList(
+        id: String,
+        elements: List<RowWithCheckboxElement>,
+    ): BottomSheet = this.apply {
+        val listContent = ListWithCheckbox(id = id, elements = elements)
+        bottomSheetModel = bottomSheetModel.copy(content = bottomSheetModel.content.toMutableList().also { it.add(listContent) })
```
yamal-alm, line 113: WDYT about having a temporary mutable list where we add these elements and build the `BottomSheetModel` when `show` method is called? (That mutable list would be cleared after that)
That way, we expand the list more eficiently

library/src/main/java/com/telefonica/mistica/bottomsheet/BottomSheetModel.kt
gmerinojimenez: About the list of lists:
[The idea is to have a list of components that can be added to the bottom sheet](https://confluence.tid.es/display/CTO/%5BAPPS%5D%5BDraft%5D+Shared+Spec%3A+Bottom+Sheet+opened+from+Webview-Bridge). Maybe we're being too generic here, but it's the way to support whatever the bridge wants to send to us

About the sealed class:
Yes, my initial idea was to have Row elements, but I refactored it to a less generic implementation. I will remove the sealed class and flat that model :)

library/src/main/java/com/telefonica/mistica/bottomsheet/BottomSheet.kt
gmerinojimenez: Keep in mind that this list is the list of components that the bottom sheet will have (a component may be a list(the only one implemented in this first iteration) / a group of chips / a group of buttons. So the size will be equal or less than **3**. So I think there will not be a performance issue and I think in this way is more clear (just having an immutable model that represents the entire bottom sheet). WDYT?

yamal-alm: you are right! performance won't be a problem here

library/src/main/java/com/telefonica/mistica/bottomsheet/BottomSheetModel.kt
yamal-alm: I see 👍

yamal-alm (approved):
very nice job 👏

library/src/main/java/com/telefonica/mistica/bottomsheet/README.md
```diff
@@ -0,0 +1,35 @@
+# Bottom Sheet
+
+Badges are placed on an overlay, at the top right side of any other view:
```
dpastor, line 3: Wrong description :D

library/src/main/res/values/styles_bottom_sheet_menu.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<style name="BottomSheetWithBackground" parent="@style/Widget.Design.BottomSheet.Modal">
+		<item name="android:background">@drawable/bottom_sheet_background</item>
```
dpastor, line 4: `bottom_sheet_background` access `?colorBackground` which won't be defined in this style. It will take parent style definition for that attribute, instead of mistica definition, no?

library/src/main/res/values/styles_bottom_sheet_menu.xml
```diff
@@ -0,0 +1,14 @@
...
+	</style>
+	<style name="BaseBottomSheetDialog" parent="MisticaTheme">
```
dpastor, line 6: `MisticaTheme` is kind of "abstract" theme, is only intended to be inherited by specific brands to assign specific brands colors (like `MisticaTheme.Telefonica`), in this case an incomplete theme definition would be used. For what is it needed? If there's anything needed from that style, could it be copied here without inheritance?

PR #187, opened 2022-10-24
ANDROID-11232 Adding ripple to actions

gmerinojimenez:
Adding the ripple effect like the one we have in rows into Actions.

Reusing the same drawable as the rows, since I assume it will always be the same.

https://user-images.githubusercontent.com/4595241/197482568-0c98a292-14a5-4d5e-b90c-98d7915f6f13.mp4


jeslat (approved):

dagonco (approved):
👨🏼‍🎨🚀

PR #189, opened 2022-11-03
ANDROID-11216 update coil version

jdelga:
Update coil version as it depends on a version of okhttp that has a vulnerabilty.

Update coil version:The minimum supported API is now 21.
No breaking changes apply: https://coil-kt.github.io/coil/upgrading/
Coil 2.2.0 requires update compile sdk to 32 as it depends on compose 1.2.1: https://coil-kt.github.io/coil/changelog/ so version 2.1.0 has been applied

### ☑️ Checks
- [X] I updated the documentation (readmes, wikis, etc). No components are changed
- [X] Tested with dark mode. No components changed
- [X] Tested with API 21: Coil version minsdk is 21

Tested using catalog app

### 📓 Notes

After reviewing further version updates, we are blocked by current sdk version. I have created a newer task to increase versions and use a compose a stable version: https://jira.tid.es/browse/ANDROID-11299

### Poyake
kotlin-extensions are not used in the project

### DIFF
--- a/catalog-compose/build.gradle
+++ b/catalog-compose/build.gradle
@@ -51,8 +51,7 @@ dependencies {
     implementation "com.google.android.material:compose-theme-adapter:1.1.5"
     implementation "androidx.navigation:navigation-compose:2.4.1"
     implementation "com.google.accompanist:accompanist-pager:$accompanist_version"
-    implementation "io.coil-kt:coil:1.4.0"
-    implementation "io.coil-kt:coil-compose:1.4.0"
+    implementation "io.coil-kt:coil-compose:$coil_version"
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.3'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
--- a/catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
+++ b/catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
@@ -16,9 +16,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.unit.dp
+import coil.compose.rememberAsyncImagePainter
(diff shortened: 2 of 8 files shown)


catalog-compose/build.gradle
```diff
@@ -51,8 +51,7 @@ dependencies {
     implementation "com.google.android.material:compose-theme-adapter:1.1.5"
     implementation "androidx.navigation:navigation-compose:2.4.1"
     implementation "com.google.accompanist:accompanist-pager:$accompanist_version"
-    implementation "io.coil-kt:coil:1.4.0"
-    implementation "io.coil-kt:coil-compose:1.4.0"
+    implementation "io.coil-kt:coil-compose:$coil_version"
```
dagonco, line 54: Apart from version upgrading, I have seen that in the Jetpack Compose section there are a couple of changes described that deprecate methods from v1 to use others in v2.

Sorry if you've already checked and they don't affect us. Just mention it to be sure.

jdelga: Actually we only use two methods from Coil and they are not affected

gmerinojimenez: Check https://coil-kt.github.io/coil/upgrading/#jetpack-compose, there are suggested changes in the compose implementation, as @dagonco points

jdelga: You are right guys, sorry

catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
```diff
@@ -16,9 +16,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.unit.dp
+import coil.compose.rememberAsyncImagePainter
 import coil.compose.rememberImagePainter
```
gmerinojimenez, line 23: unused?

gmerinojimenez (approved):
Thanks! good job (please check the comment)

catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/Lists.kt
jdelga: Thanks, fixed

dagonco (approved):
Marvelous job!

PR #192, opened 2022-11-22
ANDROID-11369 snackbar allow 10 seconds

jdelga:
Allow snackbar of 5 or 10 seconds if it does not have an action

* Add component property
* Add catalog option

### ☑️ Checks
- [X] Tested with dark mode.
- [X] Tested with API 21.

It can be tested in the dialog app

[mistica_10_result.webm](https://user-images.githubusercontent.com/36664452/203350341-802e3c61-a674-46e1-9229-6fc3bccc295f.webm)

### DIFF
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/fragment/SnackBarCatalogFragment.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/fragment/SnackBarCatalogFragment.kt
@@ -55,8 +55,8 @@ class SnackBarCatalogFragment : Fragment() {
                     }
                 }
                 val duration = when {
-                    snackbarLenght10.isChecked -> SnackbarLenght.LENGTH_10
-                    else -> SnackbarLenght.LENGTH_5
+                    snackbarLenght10.isChecked -> SnackbarLenght.LONG
+                    else -> SnackbarLenght.SHORT
                 }
                 when (SnackBarType.valueOf(dropDownInput.dropDown.text.toString())) {
                     SnackBarType.INFORMATIVE -> showInformative(duration)
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/fragment/SnackBarCatalogFragment.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/fragment/SnackBarCatalogFragment.kt
@@ -7,9 +7,11 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.inputmethod.InputMethodManager
 import android.widget.Button
+import android.widget.RadioButton
 import androidx.fragment.app.Fragment
 import com.telefonica.mistica.catalog.R
 import com.telefonica.mistica.feedback.SnackbarBuilder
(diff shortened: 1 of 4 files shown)


catalog/src/main/java/com/telefonica/mistica/catalog/ui/fragment/SnackBarCatalogFragment.kt
```diff
@@ -48,12 +51,16 @@ class SnackBarCatalogFragment : Fragment() {
             SnackbarBuilder(view, inputText.text.toString()).apply {
                 inputAction.text.toString().let { actionText ->
                     if (actionText.isNotEmpty()) {
-                        withAction(actionText, View.OnClickListener { })
+                        withAction(actionText, { })
                     }
                 }
+                val duration = when {
+                    snackbarLenght10.isChecked -> SnackbarLenght.LENGTH_10
+                    else -> SnackbarLenght.LENGTH_5
```
gmerinojimenez, line 59: minor, but I would use SHORT and LONG, similar to for example how Toast does, so the name is decoupled from the duration (if in the future we want that SHORT takes 6 seconds we will not need to replace all uses)

gmerinojimenez (approved):
So fast! thanks!

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
```diff
@@ -111,9 +110,23 @@ open class SnackbarBuilder(view: View?, text: String) {
         action.isAllCaps = false
     }
 
+    companion object {
+        private const val MAX_TEXT_LINES = 4
+    }
+}
+
+enum class SnackbarLenght {
```
dagonco, line 118: Got a type there!

library/src/main/java/com/telefonica/mistica/feedback/README.md
```diff
@@ -21,6 +21,8 @@ Builder allows SnackBar customization:
 * `withCallback(Callback callback)`
   * Adds a callback for dismiss action. Dismiss action by definition will only work when using a coordinator layout as anchor view for the SnackBar.
 
-Finally, depending on the type of SnackBar, use one of the following to display it:
+Finally, depending on the type of SnackBar, use one of the following to display it. These methods allow an argument to choose the duration of 5 second or 10.
+5 is the default.
 * `showInformative()`
+* `showInformative(SnackbarLenght.LENGTH_10)`
```
dagonco, line 27: According the code, the valid values are SHORT and LONG. Perhaps I would detail both in the documentation.

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
jdelga: Thanks! I introduced it all over the place!

library/src/main/java/com/telefonica/mistica/feedback/README.md
jdelga: Done

PR #194, opened 2022-12-14
ANDROID-11446 Deleting Mistica-compose library

gmerinojimenez:
Deleting Mistica library and including all the composables inside `mistica` library
Mostly moving packages and deleting the mistica-library publication code

### ☑️ Checks


library/src/main/java/com/telefonica/mistica/compose/README.md
```diff
@@ -117,7 +117,7 @@ There is a Mística-compose catalog in the [Demo app](https://github.com/Telefon
 | Title                     | ✅  | Composable    |
 
 ***Note**: Not every text input type have been migrated, see [text input](https://github
-.com/Telefonica/mistica-android/tree/master/library-compose/src/main/java/com/telefonica/mistica/compose/input/README.md) section to see which ones are migrated.
+.com/Telefonica/mistica-android/tree/master/library/src/main/java/com/telefonica/mistica/compose/input/README.md) section to see which ones are migrated.
```
dagonco, line 120: Here we're setting `master` branch. Probably, if someday change to `main` there is a chance that we could forgot rename it.

For me, seems like a nice idea to be agnostic to the main branch. WDYT?

dagonco (approved):
Great @gmerinojimenez! Nice job!

library/src/main/java/com/telefonica/mistica/compose/README.md
gmerinojimenez: totally, I'll change it

yamal-alm (approved):
thanks a lot 👏

library/src/main/java/com/telefonica/mistica/compose/README.md
dagonco: I don't know why but I thought it wasn't necessary... 🤦🏻‍♂️

Using first option the link wont be broken but "outdated". So master is nice, @gmerinojimenez no need to do anything to this file. Sorry!

gmerinojimenez: anyway, you had a point, thanks both of you for the info :)

PR #195, opened 2022-12-15
ANDROID-11447 Creating AbstractMisticaComposeView and porting Carousel to Views

gmerinojimenez:
[ANDROID-11447](https://jira.tid.es/browse/ANDROID-11447)

Preparing Mistica to have ComposeViews. To demonstrate the way of working we've ported Carousel and CarouselPageIndicator to views.

I've created an abstract base class called `AbstractMisticaComposeView` that contains the logic to know which xml brand theme is being applied and configure the compose's `MisticaTheme` accordingly. 

In the strange case where we want to have a View with a custom theme (like the one from Solar360), we can do it by setting customBrand value.

Finally, `CarouselView` and `CarouselPageIndicator` are mostly self-explanatory examples of how it works. 

### ☑️ Checks
- [X] Tested with dark mode.
- [ ] Tested with API 21.

https://user-images.githubusercontent.com/4595241/207855191-0f446861-59e0-4c53-8b8b-d049e678052d.mp4


catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/CatalogMainActivity.kt
```diff
@@ -246,6 +246,7 @@ fun Catalog(
             navigation = NavigationRoutes.CALLOUTS
         ),
     )
+        .sortedBy { it.name }
```
gmerinojimenez, line 249: unrelated, but it was becoming very messy

image

library/src/main/java/com/telefonica/mistica/carousel/CarouselPageIndicatorView.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.util.AttributeSet
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import com.telefonica.mistica.compose.carousel.CarouselPagerIndicator
+import com.telefonica.mistica.compose.carousel.CarouselState
+import com.telefonica.mistica.compose.composeview.AbstractMisticaComposeView
+
+class CarouselPageIndicatorView @JvmOverloads constructor(
```
gmerinojimenez, line 13: The PageIndicator are the dots below the Carousel of content
image

library/src/main/res/values/attrs_components.xml
```diff
@@ -281,4 +281,16 @@
...
+        <attr name="composeBrand" format="enum">
+            <enum name="movistar" value="0" />
+            <enum name="movistar_prominent" value="1" />
+            <enum name="o2" value="2" />
+            <enum name="o2_classic" value="3" />
+            <enum name="vivo" value="4" />
+            <enum name="telefonica" value="5" />
+            <enum name="blau" value="6" />
+        </attr>
+    </declare-styleable>
```
gmerinojimenez, line 295: This is the trick to know which Compose Brand theme should be applied on runtime

yamal-alm: very smart!

library/src/main/java/com/telefonica/mistica/carousel/CarouselView.kt
```diff
@@ -0,0 +1,48 @@
...
+    attrs,
+    defStyleAttr,
+) {
+
+    private lateinit var carouselState: CarouselState
+    private var itemCount: Int = 0
+    private lateinit var body: @Composable (Int) -> Unit
+
+    fun setState(carouselState: CarouselState): CarouselView = this.apply {
+        this.carouselState = carouselState
```
yamal-alm, line 25: is the view being recomposed if any of this attributes is updated?

library/src/main/java/com/telefonica/mistica/compose/composeview/AbstractMisticaComposeView.kt
```diff
@@ -0,0 +1,89 @@
...
+        BRAND_VALUE_O2,
+        BRAND_VALUE_O2_CLASSIC,
+        BRAND_VALUE_VIVO,
+        BRAND_VALUE_TELEFONICA,
+        BRAND_VALUE_BLAU,
+    )
+    annotation class BrandConfig
+
+    @BrandConfig
+    var brandConfig: Int = BRAND_VALUE_TELEFONICA
```
yamal-alm, line 50: can this be private?

yamal-alm (approved):
looks very good 👏

library/src/main/java/com/telefonica/mistica/carousel/CarouselView.kt
gmerinojimenez: it does, in fact in the Carousel example there are two ComposeViews (the carousel and the page indicator) that are sharing the same carouselState and they're updating correctly

library/src/main/java/com/telefonica/mistica/compose/composeview/AbstractMisticaComposeView.kt
gmerinojimenez: yep, good catch

library/src/main/java/com/telefonica/mistica/compose/composeview/AbstractMisticaComposeView.kt
```diff
@@ -0,0 +1,89 @@
...
+        BRAND_VALUE_VIVO,
+        BRAND_VALUE_TELEFONICA,
+        BRAND_VALUE_BLAU,
+    )
+    annotation class BrandConfig
+
+    @BrandConfig
+    private var brandConfig: Int = BRAND_VALUE_TELEFONICA
+
+    var customBrand: Brand? = null
```
dpastor, line 52: I'm thinking how we could provide this from xml or configure globally in some way, so we don't need to specify this each time a view is needed when a custom brand is provided, but i'm not sure of best approach for this.

What about some kind of a extendable map provided from a mistica object that holds mapping pairs (compose brand and xml theme)? (Currently brand objects define mapping to xml theme), probably this mapping should be **bidirectionally** placed somewhere else (and removed also from brand objects) ?

In any case we can probably do this on a next iteration.

dpastor (approved):
👍

library/src/main/java/com/telefonica/mistica/compose/composeview/AbstractMisticaComposeView.kt
gmerinojimenez: hmm keep in mind that the compose theme is already set via xml, but you're right that we need a way to do it automatically for the custom themes. I like your idea of having an object that sets it for all the views. The only think I don't understand is why is needed to keep a map?

gmerinojimenez: As discussed via chat, I'm adding a MisticaComposeCustomBrand to set the custom brand for all the composeViews

PR #196, opened 2022-12-15
ANDROID-11299 update compose libraries

jdelga:
https://jira.tid.es/browse/ANDROID-11299
* Update Compose version
* Update Compile SDK to 33
* Update Android Gradle Plugin to 7.3.1

### ☑️ Checks
- [X] Tested with dark mode.
- [X] Tested with API 21.

- [ ] 🖼️ Screenshots/Videos
- [X] Tested using the sample app. Everything works as expected


library/build.gradle
```diff
@@ -58,11 +58,11 @@ dependencies {
     api 'androidx.constraintlayout:constraintlayout-solver:2.0.4'
     api 'androidx.recyclerview:recyclerview:1.1.0'
 
-    implementation 'androidx.core:core-ktx:1.7.0'
+    implementation 'androidx.core:core-ktx:1.9.0'
     implementation "androidx.compose.ui:ui:$compose_version"
     implementation "androidx.compose.material:material:$compose_version"
     implementation "androidx.compose.ui:ui-tooling-preview:$compose_version"
-    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.4.1'
+    api "androidx.compose.runtime:runtime:$compose_version"
```
dpastor, line 65: as api?

dpastor (approved):
👍

library/build.gradle
jdelga: runtime dependency is required since the compiler version is not linked to compose version itself, so it needs to be exposed to the app module. 

Probably we should review the other dependencies as they are required for the application that uses them.

catalog/src/main/res/layout/activity_main_catalog.xml
```diff
@@ -16,6 +16,7 @@
         android:layout_marginTop="36dp"
         tools:ignore="ContentDescription"
         android:paddingStart="16dp"
+		android:paddingEnd="16dp"
```
DevPabloGarcia, line 19: Remove extra space

DevPabloGarcia (approved):
Nice! minor comment

library/build.gradle
DevPabloGarcia: I don't understand this change, can we review in a call?

PR #201, opened 2023-01-04
ANDROID-11516 remove jetifier

jdelga:
Remove jetifier

> Task :app:checkJetifier
Project ':app' does not use any legacy support libraries. If this is the case for all other projects, you can disable Jetifier by setting android.enableJetifier=false in gradle.properties.

> Task :catalog-compose:checkJetifier
Project ':catalog-compose' does not use any legacy support libraries. If this is the case for all other projects, you can disable Jetifier by setting android.enableJetifier=false in gradle.properties.

> Task :catalog:checkJetifier
Project ':catalog' does not use any legacy support libraries. If this is the case for all other projects, you can disable Jetifier by setting android.enableJetifier=false in gradle.properties.

> Task :library:checkJetifier
Project ':library' does not use any legacy support libraries. If this is the case for all other projects, you can disable Jetifier by setting android.enableJetifier=false in gradle.properties.

_Provide a description of the implementation. A list of steps would be ideal._
* _Step 1_
* _Step 2_
* _Step 3_

### ☑️ Checks
- [X] No documentation changes
- [X] Tested with dark mode.
- [ ] Tested with API 21.

Tested executing succesfully the app

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,8 +14,6 @@ org.gradle.jvmargs=-Xmx6656m -Dfile.encoding=UTF-8
 org.gradle.parallel=true
 
 android.useAndroidX=true
-# Automatically convert third-party libraries to use AndroidX
-android.enableJetifier=true
 
 kapt.incremental.apt=true
 


dagonco (approved):
Nice! 👌🏻

PR #203, opened 2023-02-01
fix: catalog app version code

jdelga:
Fix catalog app version code

After January 31st, the version code is causing an error:
Cause: android.defaultConfig.versionCode is set to -2145093261, but it should be a positive integer.
See https://developer.android.com/studio/publish/versioning#appversioning for more information.

Use time in milliseconds

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,7 +23,7 @@ android {
 }
 
 static def versionCodeDate() {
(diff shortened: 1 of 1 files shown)


yamal-alm (approved):
thanks 👏

dagonco (approved):
Go!

PR #204, opened 2023-02-14
ANDROID-11566 hide popover without animation

jdelga:
Allow `popover` component to be hidden without animation

Allow `animated` parameter which defaults to `true` to use an animation or not when `Popover` component is hidden
This behaviour allows it to he hidden from outside the component, for example, when initializing a recycled view

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

[device-2023-02-06-165822.webm](https://user-images.githubusercontent.com/36664452/218767018-c49a2a2d-ceff-45f7-a3aa-cdadf5e06286.webm)

### DIFF
--- a/catalog/src/main/res/layout/screen_popovers_catalog.xml
+++ b/catalog/src/main/res/layout/screen_popovers_catalog.xml
@@ -49,6 +49,13 @@
         android:text="Test" />
 
     <com.telefonica.mistica.button.Button
+        android:id="@+id/button_hide_no_animation"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
+        android:text="Hide no animation" />
+
+    <com.telefonica.mistica.button.Button
         android:id="@+id/button_create_popover_on_top"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
--- a/library/src/main/java/com/telefonica/mistica/feedback/popover/PopOver.kt
+++ b/library/src/main/java/com/telefonica/mistica/feedback/popover/PopOver.kt
@@ -58,18 +58,24 @@ open class PopOver(
     open fun setDismissButtonContentDescription(contentDescription: String): PopOver =
         apply { popOverData = popOverData.copy(dismissButtonContentDescription = contentDescription) }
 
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/telefonica/mistica/feedback/popover/PopOver.kt
```diff
@@ -58,18 +58,24 @@ open class PopOver(
...
+            animator.addListener(object : AnimatorListenerAdapter() {
+                override fun onAnimationEnd(animation: Animator) {
+                    popOverView?.removeWithoutAnimation()
+                    popOverView = null
+                }
+            })
+            animator.start()
+        } else {
+            popOverView?.removeWithoutAnimation()
+            popOverView = null
```
yamal-alm, line 77: I would wrap these two lines inside some `hidePopOver` or something like that and reuse it in both animated and non animated branches

jdelga: Thanks! Done

catalog/src/main/res/layout/screen_popovers_catalog.xml
```diff
@@ -48,6 +48,13 @@
         android:layout_marginTop="12dp"
         android:text="Test" />
 
+    <com.telefonica.mistica.button.Button
+        android:id="@+id/button_hide_no_animation"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
+        android:text="Hide no animation" />
```
yamal-alm, line 56: minor, I would add some checkbox or another button to showcase the animated scenario

yamal-alm (approved):
👍

catalog/src/main/res/layout/screen_popovers_catalog.xml
jdelga: The animated scenario is used in the cross to close the popover, you can check it in the video

yamal-alm: you are right! sorry for the noise

jeslat (approved):
👍

PR #208, opened 2023-02-24
ANDROID-11388 button compose

jdelga:
Provide a `Button` implementation in Compose

Provide a `Button` implementation in Compose

### ☑️ Checks
- [X] Tested with dark mode.
- [X] Tested with API 21.

Mistica Catalog

| drawing      | drawing |


library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +44,94 @@ fun Button(
...
 ) {
-    fun View.updateButton() {
-        findViewById<com.telefonica.mistica.button.ProgressButton>(R.id.button).apply {
-            setText(text)
-            setLoadingText(loadingText)
-            setIsLoading(isLoading)
-            isEnabled = enabled
-            setOnClickListener { onClickListener() }
+
+    val iconSpacing = 10.dp
```
yamal-alm, line 51: I would move this variable out to a constant :)

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +44,94 @@ fun Button(
...
+            modifier = modifier
+                .defaultMinSize(size.minWidth, size.minHeight)
+                .onGloballyPositioned {
+                    if (originalWidth == null) {
+                        originalWidth = with(density) { it.size.width.toDp() }
+                    }
+                }
+                .height(size.minHeight)
+                .applyWidth(originalWidth),
+            contentPadding = PaddingValues(16.dp, 0.dp, 16.dp, 0.dp),
```
yamal-alm, line 71: minor: if you want you can use `PaddingValues(horizontal = 16.dp, vertical = 0.dp)` instead

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +44,94 @@ fun Button(
...
+                            color = textColor,
+                            strokeWidth = size.progressBarStroke,
+                        )
+                        Spacer(modifier = Modifier.width(iconSpacing))
+                        Text(
+                            modifier = Modifier.align(Alignment.CenterVertically),
+                            text = loadingText,
+                            color = textColor,
+                            style = size.textStyle
+                        )
```
yamal-alm, line 99: I think we should add the spacer and the text only when `loadingText` is not null or blank to avoid having the circular progress slightly moved to the left

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +44,94 @@ fun Button(
...
-            setOnClickListener { onClickListener() }
+
+    val iconSpacing = 10.dp
+    val density = LocalDensity.current
+
+    val style = buttonStyle.getButtonStyleCompose()
+    val size = buttonStyle.getButtonSizeCompose()
+    val textColor = if (enabled) style.textColor else style.disabledTextColor
+
+    var originalWidth: Dp? by remember { mutableStateOf(null) }
```
yamal-alm, line 58: just curiosity: what is this for?

library/src/main/java/com/telefonica/mistica/compose/button/ButtonStyles.kt
```diff
@@ -0,0 +1,134 @@
...
+    ButtonStyle.DANGER_SMALL -> danger()
+    ButtonStyle.LINK -> link(colorResource(R.color.text_button_link_selector))
+    ButtonStyle.LINK_INVERSE -> linkInverse()
+}
+
+@Composable
+private fun primary(
+    buttonColors: ButtonColors = ButtonDefaults.buttonColors(
+        backgroundColor = colorResource(R.color.button_primary_background_selector),
+        disabledBackgroundColor = colorResource(R.color.button_primary_background_selector).disabled()
```
yamal-alm, line 58: Can we use colors from `MisticaTheme.colors` instead of xml ones?

yamal-alm (changes requested):
Very nice job 👏  just a few minor comments

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: Constants declare with `const` are only allowed for primitives and strings

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: This takes the final width of the button so it is used to keep the size when the `loadingText` is applied. Without this, the width of the button would change on recomposition

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
yamal-alm: we can declare the constant as an int and add `.dp` part later, but no worries, it is just a small optimization

dpastor:
Could you check if we can replace current implementation with a `ComposeView` that feeds from composable?

I'm adding @gmerinojimenez to the review also. He has tested it in the past and may have relevant feedback also.

jdelga:
I think it still makes sense to keep the xml implementation due to all usages that it has currently across projects

jdelga:
I think for now we can keep the xml implementation and keep working in the new Composable components, due to all usages that the xml button has.

Maybe some day as part of other task we can review its implementation and see if it worths removing the "leaking" `MaterialButton` interface, which would mean a mayor version including breaking changes.

yceballost (changes requested):
i need a build of compose version to review this pr

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +46,95 @@ fun Button(
...
-    ButtonStyle.LINK_INVERSE -> R.layout.button_link_inverse_wrapper
-    ButtonStyle.LINK -> R.layout.button_link_wrapper
-    ButtonStyle.PRIMARY_INVERSE -> R.layout.button_primary_inverse_wrapper
-    ButtonStyle.PRIMARY_SMALL_INVERSE -> R.layout.button_primary_small_inverse_wrapper
-    ButtonStyle.PRIMARY_SMALL -> R.layout.button_primary_small_wrapper
-    ButtonStyle.PRIMARY -> R.layout.button_primary_wrapper
-    ButtonStyle.SECONDARY_INVERSE -> R.layout.button_secondary_inverse_wrapper
-    ButtonStyle.SECONDARY_SMALL_INVERSE -> R.layout.button_secondary_small_inverse_wrapper
-    ButtonStyle.SECONDARY_SMALL -> R.layout.button_secondary_small_wrapper
-    ButtonStyle.SECONDARY -> R.layout.button_secondary_wrapper
```
dpastor, line 53: all these layouts can be deleted, no?

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +46,95 @@ fun Button(
...
+    val density = LocalDensity.current
+
+    val style = buttonStyle.getButtonStyleCompose()
+    val size = buttonStyle.getButtonSizeCompose()
+    val textColor = if (enabled) style.textColor else style.disabledTextColor
+
+    var originalWidth: Dp? by remember { mutableStateOf(null) }
+
+    CompositionLocalProvider(LocalRippleTheme provides style.rippleTheme) {
+        androidx.compose.material.Button(
```
dpastor, line 62: are we applying corner radius of 4.dp somewhere? as specs mention?

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
```diff
@@ -0,0 +1,50 @@
...
+        ButtonStyle.LINK,
+        ButtonStyle.LINK_INVERSE,
+        -> getSmallButtonSizeConfig()
+        else -> getDefaultButtonSizeConfig()
+    }
+
+@Composable
+private fun getDefaultButtonSizeConfig() = ButtonSizeConfig(
+    textStyle = MisticaTheme.typography.preset3Medium,
+    minHeight = 48.dp,
```
dpastor, line 35: this is the "height", not the "minHeight", right?

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -1,12 +1,43 @@
...
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
 import com.telefonica.mistica.R
 
+private val iconSpacing = 10.dp
+
 @Composable
 fun Button(
```
dpastor, line 42: Specs mention additional buttons tipologies, like IconButton and TextLink --> https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=238%3A6756&t=XA0vdHDzO9DdQY3n-0

I think we should also give support for them through specific composable function.

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +46,95 @@ fun Button(
...
+            colors = style.buttonColors,
+            border = if (enabled) style.border else style.disabledBorder,
+            elevation = ButtonDefaults.elevation(0.dp, 0.dp, 0.dp, 0.dp, 0.dp)
+        ) {
+            Box(contentAlignment = Alignment.Center) {
+                androidx.compose.animation.AnimatedVisibility(
+                    modifier = Modifier.fillMaxHeight(),
+                    visible = isLoading,
+                    enter = slideInVertically { it },
+                    exit = slideOutVertically { it },
```
dpastor, line 84: Animation has an specific pattern defined in specs --> https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0%3A1039&t=XA0vdHDzO9DdQY3n-0

We should configure it accordingly

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +46,95 @@ fun Button(
...
+            modifier = modifier
+                .defaultMinSize(size.minWidth, size.minHeight)
+                .onGloballyPositioned {
+                    if (originalWidth == null) {
+                        originalWidth = with(density) { it.size.width.toDp() }
+                    }
+                }
+                .height(size.minHeight)
+                .applyWidth(originalWidth),
+            contentPadding = PaddingValues(horizontal = 16.dp, vertical = 0.dp),
```
dpastor, line 72: Component specs define specific vertical padding of 12.dp also --> https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0%3A655&t=XA0vdHDzO9DdQY3n-0

is the button already applying that?

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
```diff
@@ -0,0 +1,50 @@
...
+        ButtonStyle.SECONDARY_SMALL,
+        ButtonStyle.LINK,
+        ButtonStyle.LINK_INVERSE,
+        -> getSmallButtonSizeConfig()
+        else -> getDefaultButtonSizeConfig()
+    }
+
+@Composable
+private fun getDefaultButtonSizeConfig() = ButtonSizeConfig(
+    textStyle = MisticaTheme.typography.preset3Medium,
```
dpastor, line 34: should we use medium presets, or regular?

Specs mention regular weights -> https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0%3A719&t=XA0vdHDzO9DdQY3n-0

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: I don't understand what animation do you mean. Video links are not working

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
jdelga: XML button was updated to this specs ? Do I need to check it too ?

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +46,95 @@ fun Button(
...
+                                painterResource(id = it),
+                                null,
+                                modifier = Modifier
+                                    .size(size.iconSize)
+                                    .align(Alignment.CenterVertically),
+                                colorFilter = ColorFilter.tint(style.textColor)
+                            )
+                            Spacer(modifier = Modifier.width(iconSpacing))
+                        }
+                        Text(
```
dpastor, line 123: Is the ellipsis behaviour applied by default?

https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=238%3A6303&t=XA0vdHDzO9DdQY3n-0

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
dpastor: There you can see animation behaviour specified, xml component `ProgressButton` has that animation curve already implemented, you can check it on the component itself.
Please, ask design to fix links on spec also.

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
dpastor: If there are discrepancies in specs vs what is implemented, we need to find what is the correct thing. And update whatever is wrong :)

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: I don't see any difference between the ProgressButton and the new Compose button, if you can explain a bit better, please.

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: `ButtonStyle.LINK` is already implemented, I will check Icon Button 👌

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: You can check the default shape:
image

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
dpastor: `TextLink` is not the same as a button with link style. `TextLink` is an underscored clickable text.

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
dpastor: What i mean --> Is compose button by default using same opacity and bezier curve defined in the specs?

Captura de pantalla 2023-02-28 a las 12 34 22

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
jdelga: Docs show `preset3Medium` and `preset2Medium` for small

https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0%3A1039&t=15OzfbUIgHqcvfYm-0

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
jdelga: Renamed

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
dpastor: Ok, when clicking on button details shows medium, right.
I just saw text styles page, which does not explicitly mention "medium" weight (and regular is the default) -->

Captura de pantalla 2023-03-01 a las 10 47 15

It looks medium is and was the correct weight.

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: done

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: done

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: done

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +50,99 @@ fun Button(
...
-private fun getButtonStyle(style: ButtonStyle) = when (style) {
-    ButtonStyle.DANGER -> R.layout.button_danger_wrapper
-    ButtonStyle.DANGER_SMALL -> R.layout.button_danger_small_wrapper
-    ButtonStyle.LINK_INVERSE -> R.layout.button_link_inverse_wrapper
-    ButtonStyle.LINK -> R.layout.button_link_wrapper
-    ButtonStyle.PRIMARY_INVERSE -> R.layout.button_primary_inverse_wrapper
-    ButtonStyle.PRIMARY_SMALL_INVERSE -> R.layout.button_primary_small_inverse_wrapper
-    ButtonStyle.PRIMARY_SMALL -> R.layout.button_primary_small_wrapper
-    ButtonStyle.PRIMARY -> R.layout.button_primary_wrapper
-    ButtonStyle.SECONDARY_INVERSE -> R.layout.button_secondary_inverse_wrapper
```
yceballost, line 50: we changed hover of buttonSecondary months ago, would be nice to have the last definition here <3

image

https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0%3A256&t=xucSY1ydSOKnBETg-4

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -15,43 +50,99 @@ fun Button(
...
-    AndroidView(
-        modifier = modifier,
-        factory = { context ->
-            val view = LayoutInflater.from(context).inflate(getButtonStyle(buttonStyle), null)
-            view.updateButton()
-            view
-        },
-        update = { view -> view.updateButton() }
-    )
 }
```
yceballost, line 143: Can you remove uppercase from mística catalog buttons examples? Buttons always have to use sentence case

yceballost (changes requested):

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
dpastor: Ok!

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
dpastor: Nicely done ;)

yamal-alm (approved):

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
jdelga: Done! Thanks!

yceballost (approved):
really nice 😍

PR #213, opened 2023-03-06
ANDROID-11317 movistar brand update

jdelga:
Update Movistar Brand: https://github.com/Telefonica/mistica-design/blob/production/tokens/movistar.json

Update colors in Movistar Theme and Brand

### ☑️ Checks
- [X] Tested with dark mode.
- [X] Tested with API 21.

Before

After


jeslat (approved):

pmartinbTEF (approved):
LGTM

jdelga:
Applied newer colors: https://github.com/Telefonica/mistica-design/pull/1109/files

yceballost (changes requested):
cards (media and data) background (backgroundContainer) is not correct in compose version
image

jdelga:
I have fixed the issue in the data card. However, this was not related to this task, but to an error in the data card implementation.

If we find an issue in production code, we should use a bug instead.

yceballost (approved):
perfect! thanks!!

PR #215, opened 2023-03-08
ANDROID-11767 Fix debug catalog posted link

jeslat:
Fix debug catalog posted link. We were not posting the link of the last generated version.

To fix it I've changed the API that we were using. Now we do two requests, in the first one we get the last uploaded version ID and in the second one we get the link to download that version.


jdelga (approved):
❤️

PR #217, opened 2023-03-17
ANDROID-11753 remove o2 classic theme

jdelga:
[ANDROID-11753](https://jira.tid.es/browse/ANDROID-11753)
Remove O2 Classic Theme. This PR contains breaking changes.

Remove all appearances of o2 classic

### ☑️ Checks
- [X] Tested with dark mode.
- [X] Tested with API 21.

Get APK Below


DevPabloGarcia (approved):
Thanks!

library/src/main/java/com/telefonica/mistica/compose/composeview/AbstractMisticaComposeView.kt
```diff
@@ -65,16 +62,15 @@ abstract class AbstractMisticaComposeView @JvmOverloads constructor(
     companion object {
         const val BRAND_VALUE_MOVISTAR = 0
         const val BRAND_VALUE_O2 = 1
-        const val BRAND_VALUE_O2_CLASSIC = 2
```
dagonco, line 68: Probably could be nice if we keep following a consecutive number order.

library/src/main/res/values/attrs_components.xml
```diff
@@ -292,7 +292,6 @@
         <attr name="composeBrand" format="enum">
             <enum name="movistar" value="0" />
```
dagonco, line 293: I dunno if this lead a problem, but it seems its this is bounded to @BrandConfig annotation. Maybe keeping same order and values it's worth.

dagonco (approved):
Nice cleaning, just some minors to check, but everything is cool! 🚀

library/src/main/res/values/attrs_components.xml
jdelga: Thanks! Good catch!

library/src/main/java/com/telefonica/mistica/compose/composeview/AbstractMisticaComposeView.kt
jdelga: Done

UPGRADING.md
```diff
@@ -4,7 +4,8 @@
...
+2.0.0 is the first version of Mistica that supports [dark mode](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme) in Android 10 (API 
+level 29) and higher. In order to do so, every Mistica theme passes from extending `Theme.MaterialComponents.Light.NoActionBar` to extend `Theme.MaterialComponents.DayNight.NoActionBar`. This means that if you're using `MisticaTheme` or any of the brand variants(`MisticaTheme.Movistar`, `MisticaTheme.O2` or `MisticaTheme.Vivo`) in your app, the moment you use the 2.x version of Mistica your app will support dark mode and automatically change the colors of the Mistica components when the device is set to dark mode.
```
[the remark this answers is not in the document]
jdelga: It seems it has applied some formatting. I have reverted it. Thanks!

PR #220, opened 2023-03-21
ANDROID-10988 Change 'Link checker' trigger on PR instead of dispatch and fix bad links

haynlo:
Trigger 'Link checker' workflow validation on Pull Request and fix bad links.

Change logic in .yml file
Fix bad links reported by workflow

link_checker


library/src/main/java/com/telefonica/mistica/compose/README.md
```diff
@@ -12,27 +12,27 @@ repositories {
...
-    ...
+    // ...
 }
 ```
 
 ## Configuration
 
 The goal of the team is to provide a fully compose implementation of the library, but currently there are still wrapped views inside `AndroidView` composables.
-So it is still needed to add [the brand theme in the Manifest](../README.md#configuration)
+So it is still needed to add [the brand theme in the Manifest](../../../../../../../../README.md#configuration)
```
dagonco, line 35: Probably [...](#configuration) it's enough to reference configuration section.

dagonco (approved):
Great job! 🚀

library/src/main/java/com/telefonica/mistica/compose/README.md
haynlo: I made a quick test and it didn't work as expected :(

If I declare only `(#configuration)` it only points to the `#configuration` above in the same file, and I want to point to the `#configuration` from the top level `README.md`

.github/workflows/link_checker.yml
```diff
@@ -11,6 +11,8 @@ jobs:
 
       - name: Link Checker
         id: lychee
-        uses: lycheeverse/lychee-action@v1.5.1
+        uses: lycheeverse/lychee-action@v1.6.1
```
jeslat, line 14: I would put just the major version number:
```
uses: lycheeverse/lychee-action@v1
```

README.md
```diff
@@ -22,20 +22,20 @@ repositories {
     mavenCentral()
 }
 dependencies {
-    ...
+    // ...
```
jeslat, line 25: If the three points are generating problems I think we can remove them. In my opinion, just the `implementation` line is enough to understand how to add the dependency.

library/src/main/java/com/telefonica/mistica/compose/README.md
```diff
@@ -12,27 +12,27 @@ repositories {
     mavenCentral()
 }
 dependencies {
-    ...
+    // ...
```
jeslat, line 15: Same here, I would remove the three points

jeslat (approved):
Good job!

PR #222, opened 2023-03-27
ANDROID-11727 Compose BOM update

jeprubio:
Last week a new version of BOM was released. I've updated the BOM version to 2023.03.00

Update the BOM version and make it compile. In order to compile I also had to update the kotlin version (error jdk8.CollectionsJDK8Kt found in modules kotlin-stdlib-1.8.10 ... and kotlin-stdlib-jdk8-1.7.20).

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

It's a library update. Just check it works the same.
- [ ] 🖼️ Screenshots/Videos
- [ ] ...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,11 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext {
-        compose_bom_version = "2023.01.00"
-        compose_compiler_version = '1.3.2'
-        kotlin_version = "1.7.20"
+        compose_bom_version = '2023.03.00'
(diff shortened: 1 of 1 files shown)


jeprubio:
It was compiling locally but I guess was due to some cache or something, closing for now.

jeprubio:
> It has failed because some lint checks with kotlin 1.8. I think we can just update compose and not kotlin.

More or less, you can update compose and not kotlin as long as you don't update the compose compiler but if I try to compile the recently released 2023.03.00 without updating kotlin and the compose compiler I can't.

PR #223, opened 2023-03-30
Adding sarif to pull requests

fjquileze:


dagonco:
You've invoked whole Android team! Adding once @telefonica-android should be capable enough to add 2 reviewers randomly.

image

.github/workflows/tests.yml
```diff
@@ -11,4 +11,11 @@ jobs:
     - name: "Build Android project"
       uses: docker://docker.tuenti.io/android/novum_android_api31:2
       with:
-        args: 'bash ./gradlew clean check assemble'
+        args: 'bash ./gradlew clean check assemble lint'
```
dagonco, line 14: I can't tell right now if linter is already being executed with `check` task. But probably it is. Can you double check that?

dagonco (approved):
Great first PR! Let's go! 🚀

gmerinojimenez (approved):
Nice work!! congrats!

gmerinojimenez:
> You've invoked whole Android team! Adding once @telefonica-android should be capable enough to add 2 reviewers randomly.

I think that was my bad 🤨

DevPabloGarcia (approved):
Nice first PR ! 🔥

PR #225, opened 2023-04-03
ANDROID-11501 load error feedback compose

jdelga:
Add Load Error Feedback Compose implementation

Add the new component
Add helpers into catalog

### ☑️ Checks
- [X] Tested with dark mode.
- [X] Tested with API 21.

image
image


library/src/main/java/com/telefonica/mistica/compose/feedback/error/LoadErrorFeedback.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.telefonica.mistica.compose.theme.brand.MovistarBrand
+
+@Composable
+fun LoadErrorFeedback(
+    titleText: String? = null,
+    descriptionText: String? = null,
+    isLoading: Boolean = false,
+    buttonText: String? = null,
+    buttonLoadingText: String = "",
+    loadingAction: () -> Unit,
```
yamal-alm, line 24: I would use something like `onClick` or `onButtonClick`

library/src/main/java/com/telefonica/mistica/compose/feedback/error/LoadErrorFeedback.kt
```diff
@@ -0,0 +1,91 @@
...
+) {
+
+    Column(modifier = Modifier.padding(16.dp)) {
+        titleText?.let {
+            Text(
+                modifier = Modifier
+                    .padding(8.dp)
+                    .align(Alignment.CenterHorizontally),
```
yamal-alm, line 32: this can be moved to the parent `Column` just one time instead of on every child:
```
Column(
   ...
   horizontalAlignment = Alignment.CenterHorizontally
)
```

library/src/main/java/com/telefonica/mistica/compose/feedback/error/README.md
```diff
@@ -0,0 +1,28 @@
...
+val scope = rememberCoroutineScope()
+var isLoading by remember { mutableStateOf(false) }
+
+LoadErrorFeedback(
+    titleText = "Title",
+    descriptionText = "Something went wrong, this is the description of the failure. Try to retry if you want.",
+    isLoading = isLoading,
+    buttonText = "Retry",
+    buttonLoadingText = "Loading"
+) {
```
yamal-alm, line 21: I would add here the name of the `loadingAction` param to make it more explicit. I think using it as a trailing lambda may lead to confusion and make the developer think that it accepts other composables WDYT

yamal-alm (changes requested):
Looks good 👏  just a few comments

catalog-compose/src/main/java/com/telefonica/mistica/compose/catalog/ui/components/LoadErrorFeedbacks.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    var isLoading by remember { mutableStateOf(false) }
+
+    Column(
+        modifier = Modifier.fillMaxSize(),
+        verticalArrangement = Arrangement.spacedBy(8.dp),
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        OutlinedTextField(
+            modifier = Modifier.padding(horizontal = 50.dp),
```
dagonco, line 44: That's a minor, probably this field width will be increased if text is long enough.
Maybe something like Modifier.fillwidth().padding(yourHorizontalPadding) could avoid that.

library/src/main/java/com/telefonica/mistica/compose/feedback/error/LoadErrorFeedback.kt
```diff
@@ -0,0 +1,91 @@
+package com.telefonica.mistica.compose.feedback
```
dagonco, line 1: Check package.
```suggestion
package com.telefonica.mistica.compose.feedback.error
```

dagonco (approved):
That's cool!

jdelga:
Applied comments

yamal-alm (approved):
👏

yceballost:
This component never was defined by Mística Design team, is just a legacy from Tuenti. So.. should we include this component in Mística?

This component can be replaced by Feedback Error Screen

jdelga:
I think we should add the compose version of this component to keep working on the new compose components. However, we can consider a new component to substitute this one or remove it.

PR #226, opened 2023-04-05
ANDROID-11911 Add public appcenter link

jeslat:
Add public appcenter link to be able to download the testing catalog if the download link has expired.

You can see the result in the comment.


dagonco (approved):
Nice! Thanks for the investigation!

yamal-alm (approved):
thanks!

PR #227, opened 2023-04-05
ANDROID-11727 Fix BOM dependencies

yamal-alm:
When trying to import latest Mistica versioN (7.3.1) on any project, compilation fails with this error:
```
* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all files for configuration ':app:movistarESDebugCompileClasspath'.
   > Could not find androidx.compose.ui:ui-tooling-preview:.
     Required by:
         project :app > com.telefonica:mistica:7.3.1
```

It looks like `ui-tooling` and `ui-tooling-preview` are two libraries that are related, using debugImplementation for both of them instead of just for `ui-tooling` seems to fix the error.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

- Not a UI change, there is no need for this change to be reviewed by any member from Design team.
- I published a snapshot version with this changes and the error is gone

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -63,13 +63,15 @@ dependencies {
     implementation 'androidx.core:core-ktx:1.9.0'
     implementation "androidx.compose.ui:ui"
     implementation "androidx.compose.material:material"
-    implementation "androidx.compose.ui:ui-tooling-preview"
     api "androidx.compose.runtime:runtime"
     implementation 'androidx.activity:activity-compose:1.4.0'
     implementation "com.google.android.material:compose-theme-adapter:1.1.5"
     implementation "com.google.accompanist:accompanist-pager:$accompanist_version"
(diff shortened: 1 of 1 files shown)


library/build.gradle
```diff
@@ -63,29 +63,23 @@ dependencies {
...
     testImplementation composeBom
     testImplementation 'junit:junit:4.13.2'
     testImplementation "androidx.compose.ui:ui-test-junit4"
     testImplementation "org.mockito.kotlin:mockito-kotlin:4.0.0"
-
-    androidTestImplementation composeBom
-    androidTestImplementation 'androidx.test.ext:junit:1.1.3'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
-    androidTestImplementation "androidx.compose.ui:ui-test-junit4"
-    androidTestImplementation("androidx.compose.ui:ui-test-manifest")
```
yamal-alm, line 86: not needed right now

PR #229, opened 2023-04-10
ANDROID-11916 Text Area in Compose

jeprubio:
### :tickets: Jira ticket
[ANDROID-11916](https://jira.tid.es/browse/ANDROID-11916)

Add a TextArea component as the one there is in the gallery that it's using traditional views.

- Created new TextAreaInput composable.
- Created new CharsCounter composable.
- Updated TextInputImpl so that it allows multiline when isTextArea
- Added underlineEnd slot to TextInputImpl so we can pass a CharsCounter when it's a textarea, this would allow any input to have CharsCounter or another composable on the bottomright if needed.

EmptyTextarea

SelectedTextarea

Lorem Ipsum

LoremIpsumTruncated

https://user-images.githubusercontent.com/13270085/230917350-d1325bd9-ea48-46ae-af3e-6bb76d0c1e15.mov

### ☑️ Checks
- [x] Tested with dark mode.
- [ ] Tested with API 21.

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
@@ -25,7 +25,7 @@ fun TextAreaInput(
     enabled: Boolean = true,
     readOnly: Boolean = false,
     onClick: (() -> Unit)? = null,
-    maxChars: Int = Int.MAX_VALUE,
+    maxChars: LimitCharacters = LimitCharacters.Unlimited,
     visualTransformation: VisualTransformation = VisualTransformation.None,
     keyboardOptions: KeyboardOptions = KeyboardOptions.Default,
 ) {
@@ -37,9 +37,7 @@ fun TextAreaInput(
         modifier = modifier,
         value = value,
         onValueChange = { newValue ->
-            val text = if (newValue.length > maxChars) newValue.substring(startIndex = 0, endIndex = maxChars) else newValue
-            currentChars = text.length
-            onValueChange(text)
+            currentChars = doOnValueChange(maxChars, newValue, onValueChange)
         },
         label = label,
         helperText = helperText,
@@ -53,13 +51,32 @@ fun TextAreaInput(
         isTextArea = true,
         visualTransformation = visualTransformation,
         underlineEnd = {
-            if (maxChars < Int.MAX_VALUE) {
+            if (maxChars is LimitCharacters.Limited) {
                 CharsCounter(current = currentChars, max = maxChars)
             }
         },
         keyboardOptions = keyboardOptions.copy(imeAction = ImeAction.None)
             .toFoundationKeyboardOptions(KeyboardType.Text),
     )
+}
+
+private fun doOnValueChange(
+    maxChars: LimitCharacters,
+    newValue: String,
+    onValueChange: (String) -> Unit,
+): Int {
+    return if (maxChars is LimitCharacters.Limited) {
+        val text = if (newValue.length > maxChars.characterLimit) {
+            newValue.substring(startIndex = 0, endIndex = maxChars.characterLimit)
+        } else {
+            newValue
+        }
+        onValueChange(text)
+        text.length
+    } else {
+        onValueChange(newValue)
+        newValue.length
+    }
 }
 
 @Preview(showBackground = true)
--- a/library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
@@ -52,7 +52,11 @@ fun TextAreaInput(
         visualTransformation = visualTransformation,
         underlineEnd = {
             if (maxChars is LimitCharacters.Limited) {
-                CharsCounter(current = currentChars, max = maxChars)
+                CharsCounter(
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/telefonica/mistica/compose/input/TextInputImpl.kt
```diff
@@ -134,8 +156,8 @@ private fun TextBox(
             errorIndicatorColor = Color.Transparent,
             errorCursorColor = MisticaTheme.colors.controlActive,
         ),
-        singleLine = true,
-        maxLines = 1,
+        singleLine = !isMultiLine,
```
yamal-alm, line 159: suggestion for simplicity: expose this flag as `singleLine` so we don't need to negate the boolean here.

library/src/main/java/com/telefonica/mistica/compose/input/TextInputImpl.kt
```diff
@@ -145,18 +167,25 @@ private fun Underline(
...
+            UnderlineTextAnimatedVisibility(
+                visible = isError && errorText != null,
+                text = errorText,
+                color = LocalTextInputColors.current.errorTextColor,
+            )
+        }
+        Spacer(modifier = Modifier.weight(1f))
+        Box(Modifier.padding(top = 4.dp, start = 14.dp, end = 14.dp)) {
+            underlineEnd?.invoke()
+        }
```
yamal-alm, line 188: I would add these two items conditionally based on the nullability of `underlineEnd` so we don't add any empty space if it's not present

library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
```diff
@@ -0,0 +1,84 @@
...
+    modifier: Modifier = Modifier,
+    helperText: String? = null,
+    isError: Boolean = false,
+    errorText: String? = null,
+    trailingIcon: @Composable (() -> Unit)? = null,
+    isInverse: Boolean = false,
+    enabled: Boolean = true,
+    readOnly: Boolean = false,
+    onClick: (() -> Unit)? = null,
+    maxChars: Int = Int.MAX_VALUE,
```
yamal-alm, line 28: Should we show `2147483647` if no `maxChars` is provided? Maybe it is something to confirm with design team but I wouldn't show any number at all.

yamal-alm (commented):
just a few comments :) I've added as well something from design to verify these changes. Thanks!

library/src/main/java/com/telefonica/mistica/compose/input/TextInputImpl.kt
jeprubio: Good one, applied, even though it is then assigned as:
```
singleLine = !isTextArea,
```

library/src/main/java/com/telefonica/mistica/compose/input/TextInputImpl.kt
jeprubio: Good shout! Applied.

library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
jeprubio: Specification like what? Can you provide any sample? Is it ok if I just use a sealed class for this or do you want something different?

dagonco: I meant if there is any Figma or document in which this behaviour is described or did you code it by your own.

yamal-alm (approved):
good job!

library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
jeprubio: I've added the interface LimitCharacters which has Unlimited and Limited, I hope it's more self explanatory.

I've seen, the problem is that this task I understood was about transforming the current text area in compose, the figma file has for example errors when there are more characters in the textarea than the allowed and that's not possible with the current implementation of the textarea. Also some margins of the helper text and the counter are slightly different. So what I did was to copy the current implementation in compose and only added something extra which is that with this implementation we would be able to add easily a chars counter to any TextInput even though now it's only used in the textarea but apart from that I copied the same exact current behaviour instead of what figma says.

dagonco: Okay, if you've copied same behaviour is nice, that's what I wanted to say. So great!

dagonco (approved):
Nice job!

library/src/main/java/com/telefonica/mistica/compose/input/TextAreaInput.kt
aweell: To understand better this part:

* The textarea maxChars is configurable, it can have a limit (n) or unlimited.
   * When `limited`: a char counter will appear showing char introduced / n, when the limit is reached it's not possible to add more characters
   * When `unlimited`: the char counter is not visible, there's no limit
   
I'm understanding it in the right way?

jeprubio: Yup, that's exactly it.

aweell (approved):

PR #230, opened 2023-04-11
Fix compose bom releases

yamal-alm:
When Compose BoM was included on Mistica (Version 7.3.1). The compilation failed on every project that didn't use Compose and Mistica dependency is added. This is the error:
```
* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all files for configuration ':app:movistarESDebugCompileClasspath'.
   > Could not find androidx.compose.ui:ui-tooling-preview:.
     Required by:
         project :app > com.telefonica:mistica:7.3.1
```

The BoM usage was temporarily reverted in #228, so I reverted the revert

I haven't find the real reason behind this but setting the explicit version on Mistica for that package does the trick. It allows the version to be included in the `pom.xml` published to Maven. 

On the other hand, we haven't seen the error occur for other packages from the BoM so I hope that this is an error that may be fixed in future.

I added a small comment in `build.gradle` referencing to this same PR to remember ourselves in the future that `ui-tooling-preview` version should be updated when the BoM is updated as well.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

- Pubished a local snapshot in maven local and tested that all dependencies are successfully resolved

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext {
-        compose_version = '1.3.1'
+        compose_bom_version = "2023.01.00"
         compose_compiler_version = '1.3.2'
         kotlin_version = "1.7.20"
         androidx_appcompat_version = "1.5.1"
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,11 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext {
-        compose_bom_version = "2023.01.00" // Remember to update ui-tooling-preview version as well.
+        // Remember to update "ui_tooling_preview_version" when updating "compose_bom_version"
+        // according to https://mvnrepository.com/artifact/androidx.compose/compose-bom/
+        // For more info see https://github.com/Telefonica/mistica-android/pull/230
+        compose_bom_version = "2023.01.00"
+        ui_tooling_preview_version = "1.3.3"
         compose_compiler_version = '1.3.2'
         kotlin_version = "1.7.20"
         androidx_appcompat_version = "1.5.1"
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext {
-        compose_bom_version = "2023.01.00"
+        compose_bom_version = "2023.01.00" // Remember to update ui-tooling-preview version as well.
         compose_compiler_version = '1.3.2'
         kotlin_version = "1.7.20"
         androidx_appcompat_version = "1.5.1"
--- a/catalog-compose/build.gradle
+++ b/catalog-compose/build.gradle
@@ -40,21 +40,20 @@ android {
 
 dependencies {
     implementation project(':library')
-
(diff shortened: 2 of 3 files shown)


library/build.gradle
```diff
@@ -63,26 +63,31 @@ dependencies {
...
+    implementation "androidx.compose.material:material"
+    api "androidx.compose.runtime:runtime"
     implementation 'androidx.activity:activity-compose:1.4.0'
     implementation "com.google.android.material:compose-theme-adapter:1.1.5"
     implementation "com.google.accompanist:accompanist-pager:$accompanist_version"
     implementation "com.google.accompanist:accompanist-pager-indicators:$accompanist_version"
     implementation "io.coil-kt:coil-compose:$coil_version"
 
+    // 1.3.3 is the version included in the currently used Compose BoM (https://mvnrepository.com/artifact/androidx.compose/compose-bom/2023.01.00)
+    implementation "androidx.compose.ui:ui-tooling-preview:1.3.3"
```
yamal-alm, line 80: this is the actual change of this PR. The other one is the revert from e83e8846dbcf99abcb32048d3ccccb4d0ec1197c

catalog-compose/build.gradle
```diff
@@ -40,21 +40,20 @@ android {
 
 dependencies {
     implementation project(':library')
-
+    implementation platform("androidx.compose:compose-bom:$compose_bom_version")
```
yamal-alm, line 43: not new changes, these changes comes from the revert

library/build.gradle
jdelga: But library uses `ui-tooling-preview` which is in the bom. Why does it need to be declared explicitly ?

build.gradle
```diff
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext {
-        compose_version = '1.3.1'
+        compose_bom_version = "2023.01.00" // Remember to update ui-tooling-preview version as well.
```
jdelga, line 4: I would define `tooling-preview` version here so they are together

yamal-alm: good idea!

yamal-alm: done 5029002b1a461fc388432d84213b56d8c8f268ba

library/build.gradle
```diff
@@ -63,26 +63,30 @@ dependencies {
     api 'androidx.constraintlayout:constraintlayout:2.0.4'
     api 'androidx.constraintlayout:constraintlayout-solver:2.0.4'
     api 'androidx.recyclerview:recyclerview:1.1.0'
+    def composeBom = platform("androidx.compose:compose-bom:$compose_bom_version")
+    api composeBom
```
jeprubio, line 67: Could this `api` be the cause of the problems with the ui-tooling-preview problems?

library/build.gradle
jeprubio: androidx.compose.ui:ui-tooling-preview is inside the bom: https://developer.android.com/jetpack/compose/bom/bom-mapping and it's 1.3.3 for 2023.01.00 🤔

library/build.gradle
yamal-alm: After trying to use `implementation` instead of `api` I got the same error, so it doesn't seem that this is the cause. Thanks anyway for the idea

jeprubio (approved):

PR #231, opened 2023-04-11
ANDROID-11931 Set default title style per brand

jeslat:
Set default title style per brand. It has to be `title2` in Movistar and `title1` in the rest of the brands.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

- [ ] 🖼️ Screenshots/Videos


library/src/main/java/com/telefonica/mistica/compose/theme/MisticaTheme.kt
```diff
@@ -98,4 +109,9 @@ object MisticaTheme {
         @Composable
         @ReadOnlyComposable
         get() = LocalMisticaTypography.current
+
+    val values: MisticaValues
```
jeslat, line 113: I've added a new `MisticaValues` class similar to `MisticaTypography` and `MisticaColors` to have values that have to be different per brand. Currently, we are going to have only `titleStyle` but in the near feature will add corner radius for example.

AnaMontes11 (approved):
Nice job!

haynlo (approved):
Good job!

dagonco (approved):
So good! 👌🏻

PR #239, opened 2023-04-21
ANDROID-11475 Add catalog migration info

haynlo:
Document the main migration catalog changes for the rest of the projects.

* Update "UPGRADING" document.

### ☑️ Checks
N/A

N/A

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,8 +1,10 @@
 # Upgrading guide
 
-## Removed 'catalog-compose' component
+## Upgrade to version 8.3.0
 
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -1,5 +1,15 @@
 # Upgrading guide
 
+## Removed 'catalog-compose' component
```
jdelga, line 3: I would add a section "Upgrade to version 8.3.0" as main title, like below, so that it is first reviewed by version

jeslat (approved):

jdelga (approved):
There is already a section "## Upgrade from 1.x to 2.x". I would put version 8 below this one

PR #244, opened 2023-05-17
ANDROID-11435 Allow custom bottom section in ListRowItem

yamal-alm:

Include the custom and optional bottom fragment in the ListRowItem according to the specs

:warning: By now it is only implemented in the Compose version of the component. It will be implemented on demand for the xml version. :warning:

Captura de Pantalla 2023-05-17 a las 12 50 20

Add a new composable called `bottom` as param to the ListRowItem composable.

Other changes in this PR:
- Set `height(IntrinsicSize.Min)` to the whole row so `trailing` content can use `fillMaxHeight`
- Fix background color for boxed rows according to the specs (`backgroundContainer `)

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [x] [Mistica App download link](https://appcenter-filemanagement-distrib5ede6f06e.azureedge.net/9752b206-3ae1-43db-bc0e-10e003abe526/MisticaCatalog-debug-1684319563.apk?sv=2019-02-02&sr=c&sig=jSzQFzC6aNlbCl%2Bhs%2BAe2lQxU51KE0yWWJBwtqL8Gnw%3D&se=2023-05-18T10%3A35%3A09Z&sp=r)
- [x] 🖼️ Screenshots/Videos

Captura de Pantalla 2023-05-17 a las 12 59 06

Captura de Pantalla 2023-05-17 a las 12 59 22

Captura de Pantalla 2023-05-17 a las 12 59 34

Captura de Pantalla 2023-05-17 a las 13 01 18

Captura de Pantalla 2023-05-17 a las 13 01 47

Captura de Pantalla 2023-05-17 a las 13 02 04

Captura de Pantalla 2023-05-17 a las 13 03 00

### DIFF
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
@@ -75,7 +75,7 @@ fun samples() = listOf(
         title = TITLE,
         subtitle = SUBTITLE,
         action = { Chevron() },
-        bottom =  { CustomFragment() },
+        bottom =  { CustomSlot() },
     ),
 
     ListItem(
@@ -116,7 +116,7 @@ fun samples() = listOf(
         subtitle = SUBTITLE,
         action = { Chevron() },
         icon = { ListIcon() },
-        bottom =  { CustomFragment() },
+        bottom =  { CustomSlot() },
     ),
 
     ListItem(
@@ -157,7 +157,7 @@ fun samples() = listOf(
         subtitle = SUBTITLE,
         action = { Chevron() },
         icon = { ListIcon() },
-        bottom =  { CustomFragment() },
+        bottom =  { CustomSlot() },
     ),
 
     ListItem(
@@ -205,7 +205,7 @@ fun samples() = listOf(
         isBadgeVisible = true,
         badge = "1",
         icon = { Avatar() },
-        bottom =  { CustomFragment() },
+        bottom =  { CustomSlot() },
     ),
 )
 
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
@@ -1,19 +1,27 @@
 package com.telefonica.mistica.catalog.ui.compose.components
 
 import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.Divider
 import androidx.compose.material.ExperimentalMaterialApi
 import androidx.compose.material.Icon
 import androidx.compose.material.Switch
+import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.painterResource
@@ -63,6 +71,12 @@ fun samples() = listOf(
         subtitle = SUBTITLE,
         action = { Chevron() },
     ),
+    ListItem(
+        title = TITLE,
+        subtitle = SUBTITLE,
+        action = { Chevron() },
+        bottom =  { CustomFragment() },
+    ),
 
     ListItem(
         title = TITLE,
@@ -97,6 +111,13 @@ fun samples() = listOf(
         action = { Chevron() },
         icon = { ListIcon() },
     ),
+    ListItem(
+        title = TITLE,
+        subtitle = SUBTITLE,
+        action = { Chevron() },
+        icon = { ListIcon() },
+        bottom =  { CustomFragment() },
+    ),
 
     ListItem(
         title = TITLE,
@@ -130,6 +151,13 @@ fun samples() = listOf(
         subtitle = SUBTITLE,
         action = { Chevron() },
         icon = { ListIcon() },
+    ),
+    ListItem(
+        title = TITLE,
+        subtitle = SUBTITLE,
+        action = { Chevron() },
+        icon = { ListIcon() },
+        bottom =  { CustomFragment() },
     ),
 
     ListItem(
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -51,15 +53,16 @@ fun ListRowItem(
     headline: Tag? = null,
     trailing: @Composable (() -> Unit)? = null,
     onClick: (() -> Unit)? = null,
+    bottom: @Composable (() -> Unit)? = null,
+    contentPadding: PaddingValues = PaddingValues(horizontal = 16.dp, vertical = 8.dp),
```
yamal-alm, line 57: This overrideable `contentPadding` allows us to not depend exclusively on the fixed padding values that we had before.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -104,7 +111,7 @@ fun ListRowItem(
         modifier = boxModifier.testTag(ListRowItemTestTags.LIST_ROW_ITEM)
     ) {
         Row(
-            modifier = rowModifier
+            modifier = rowModifier.height(IntrinsicSize.Min)
```
yamal-alm, line 114: This allows us to use `fillMaxHeight` in  `trailing` content.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -151,6 +158,10 @@ fun ListRowItem(
                             .defaultMinSize(minHeight = 20.dp),
                     )
                 }
+                bottom?.let {
+                    Spacer(modifier = Modifier.height(2.dp))
+                    bottom()
+                }
```
yamal-alm, line 164: this is the bottom custom content that was missing from the spec

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -73,6 +76,10 @@ fun ListRowItem(
                 color = MisticaTheme.colors.border,
                 shape = RoundedCornerShape(MisticaTheme.values.containerBorderRadius),
             )
+            .background(
+                color = MisticaTheme.colors.backgroundContainer,
+                shape = RoundedCornerShape(MisticaTheme.values.containerBorderRadius),
+            )
```
yamal-alm, line 82: according to the specs: `backgroundContainer` is the color that boxed rows should use

jeprubio, line 82: I think that shape = RoundedCornerShape from the border and the background would not be needed if a .clip() with that shape is done before, could you try that? I mean
```
.clip(RoundedCornerShape(MisticaTheme.values.containerBorderRadius))
.border(
    width = 1.dp,
    color = MisticaTheme.colors.border,
)
.background(
    color = MisticaTheme.colors.backgroundContainer,
)
```
I haven't tried.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
jeprubio: Another possibility would be to pass the padding in the modifier parameter each time, but I agree this looks better, nice.

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
```diff
@@ -284,3 +325,27 @@ fun Avatar(url: String) {
         modifier = Modifier.size(40.dp)
     )
 }
+
+@Composable
+private fun CustomFragment() {
```
jeprubio, line 330: Nice, I wonder if this could be internal instead of private and something that could be reused inside the catalog for different layouts, like `ColouredBox`. Just a thought.

jeprubio (commented):
Nice, I add one comment to check about using .clip() to prevent from having to define the shape in the border and the background and I would also add a preview using this bottom composable (just pass a Box with some colour to it for example).

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
yamal-alm: yes, that would be ideal. However this composable is used in many repositories and I wanted to keep retro-compatibility with these default values definded via `contentPadding` which is a pattern used in composables such as LazyColumn for instance.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
yamal-alm: After trying that, this is the result:

Captura de Pantalla 2023-05-17 a las 14 32 40

It looks like `background()` modifier uses a default `RectangularShape` that needs to be overridden (in this case by `RoundedCornerShape`) to avoid the effect from the above image

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
yamal-alm: in the future it could be a good idea 👍  but right now this is the only place where it is used and it is in the catalog so I think it's not worth it by now. Thanks anyways for the suggestion

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
jeprubio: Thanks for trying, then don't worry about that, did you manage to check if a preview could be set using this bottom composable?

yamal-alm: sure, added here ab8cdf4563222bce28d9d9c3d939383059cb96b8

Captura de Pantalla 2023-05-17 a las 15 31 50

jeprubio (approved):

yceballost:
> ### 🥅 What's the goal?
> Include the custom and optional bottom fragment in the ListRowItem according to the specs
> 
> ⚠️ By now it is only implemented in the Compose version of the component. It will be implemented on demand for the xml version. ⚠️
> 
> Captura de Pantalla 2023-05-17 a las 12 50 20
> 
> ### 🚧 How do we do it?
> Add a new composable called `bottom` as param to the ListRowItem composable.
> 
> Other changes in this PR:
> 
> * Set `height(IntrinsicSize.Min)` to the whole row so `trailing` content can use `fillMaxHeight`
> * Fix background color for boxed rows according to the specs (`backgroundContainer `)
> 
> ### ☑️ Checks
> * [x]  I updated the documentation, including readmes and wikis. If this is a breaking change, update [UPGRADING.md](../UPGRADING.md) to inform users how to proceed. If no updates are necessary, indicate so.
> * [x]  Tested with dark mode.
> * [x]  Tested with API 23.
> 
> ### 🧪 How can I test this?
> * [ ]  Reviewed by Mistica design team
> * [x]  [Mistica App download link](https://appcenter-filemanagement-distrib5ede6f06e.azureedge.net/9752b206-3ae1-43db-bc0e-10e003abe526/MisticaCatalog-debug-1684319563.apk?sv=2019-02-02&sr=c&sig=jSzQFzC6aNlbCl%2Bhs%2BAe2lQxU51KE0yWWJBwtqL8Gnw%3D&se=2023-05-18T10%3A35%3A09Z&sp=r)
> * [x]  🖼️ Screenshots/Videos
> 
> Captura de Pantalla 2023-05-17 a las 12 59 06
> 
> Captura de Pantalla 2023-05-17 a las 12 59 22
> 
> Captura de Pantalla 2023-05-17 a las 12 59 34
> 
> Captura de Pantalla 2023-05-17 a las 13 01 18
> 
> Captura de Pantalla 2023-05-17 a las 13 01 47
> 
> Captura de Pantalla 2023-05-17 a las 13 02 04
> 
> Captura de Pantalla 2023-05-17 a las 13 03 00

just to unify terminology, can you rename it to "slot" ? Fragment was the first early name for this purpose 😄

yceballost (approved):

yamal-alm:
> just to unify terminology, can you rename it to "slot" ? Fragment was the first early name for this purpose 😄

Sure, example from catalog renamed to "Custom slot": 
Captura de Pantalla 2023-05-17 a las 15 54 50

dagonco (approved):
So cool, thanks @yamal-alm !

PR #245, opened 2023-05-17
Fix generate snapshot workflow and get rid of self hosted runners

yamal-alm:
- Snapshot workflow is not working right now because there is no such task as `publishNoopPublicationToSonatypeRepository`. I'm fixing it.
- I'm also replacing `self-hosted-novum` runner with `ubuntu-latest` in some of the workflows because we actually don't need `novum_android_api31` docker image at all.

- Use the same command as in `release.yml` workflow but with `SNAPSHOT_VERSION` param
- Remove every unneeded usage of `self-hosted-novum` and use `ubuntu-latest` instead
- Remove every unneeded usage of `novum_android_api31` docker image

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- No need to be reviewed by design team
- Checked that `tests.yml` and `debug_catalog.yml` are executed successfully

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,12 +8,13 @@ concurrency:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-novum
     steps:
     - name: Checkout repo
       uses: actions/checkout@v2
     
     - name: "Build Android project"
+      uses: docker://docker.tuenti.io/android/novum_android_api31:2
       run : 'bash ./gradlew clean check assemble'
 
     - name: Upload Lint results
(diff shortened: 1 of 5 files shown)


jeslat (approved):
thanks!

.github/workflows/tests.yml
```diff
@@ -8,13 +8,15 @@ concurrency:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted-novum
```
yamal-alm, line 11: sorry, the usage of ubuntu-latest on tests was commited to master by mistake in https://github.com/Telefonica/mistica-android/pull/244. I'm reverting it. However, it isn't something critical..

jeslat (approved):

gmerinojimenez (approved):
Good job!!

PR #248, opened 2023-05-24
ANDROID-13240 compose custom icon 

jdelga:
The main goal is to add the new `CUSTOM` type to the compose implementation which was already added to the XML implementation: https://github.com/Telefonica/mistica-android/pull/243

Afterwards, I have checked there are some improvements that could be done to the catalog itself to avoid the duplication of the inputs.

### :boom: What's the goal?
* Breaking change: `TYPE_CUSTOM` must be changed to `TYPE_INFO`
* `shouldAnimateOnAttachedToWindow` change to `setShouldAnimateOnAttachedToWindow(boolean)` or by xml

* Add CUSTOM implementation in Compose
* Add `shouldAnimateOnAttached` option in the xml itself (poyake)
* Add CUSTOM icons and animations to the input options. Instead of updating the two different input layouts, I have merged them into a single one with two different launch options.

### ☑️ Checks
- [X] I updated the documentation, including readmes and wikis. If this is a breaking change, update [UPGRADING.md](../UPGRADING.md) to inform users how to proceed. If no updates are necessary, indicate so.
- [X] Tested with dark mode.
- [X] Tested with API 23.

Before:

https://github.com/Telefonica/mistica-android/assets/36664452/7bf2339b-70d2-4792-bcc4-7111e22a5b1e

After:

https://github.com/Telefonica/mistica-android/assets/36664452/b3b78f1e-1007-4604-a608-bff5dfc7e363


catalog/src/main/java/com/telefonica/mistica/catalog/ui/ComponentCatalogActivity.kt
```diff
@@ -183,6 +222,17 @@ class ComponentCatalogActivity : FragmentActivity() {
         binding.componentViewPager.adapter = ComponentPageAdapter(adapterList, this)
     }
 
+    private fun setSingleInputLauncher(
```
jdelga, line 225: New method for components that have only one input and are launched with the different implementations

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/CommonButtonLaunchBlock.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.foundation.layout.width
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import com.telefonica.mistica.compose.button.Button
+
+@Composable
+fun CommmonButtonLaunchBlock(
```
jdelga, line 18: Common launch block, to be reused

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/activity/FeedbackScreenCatalogActivity.kt
```diff
@@ -58,6 +59,10 @@ class FeedbackScreenCatalogActivity : AppCompatActivity() {
             })
             customIcon?.let { setCustomIcon(it) }
             customAnimation?.let { setCustomAnimation(it) }
+            setShouldAnimateOnAttached(shouldAnimateOnAttached)
+            if (!shouldAnimateOnAttached){
```
pmartinbTEF, line 63: shouldn't it be?
if (shouldAnimateOnAttached){

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
```diff
@@ -1,49 +1,97 @@
...
+            firstButtonText = firstButtonText,
+            firstButtonLoadingText = firstButtonLoadingText,
+            secondButtonText = secondButtonText,
+            showSecondButtonAsLink = secondButtonAsLink,
+            customAnimation = customAnimation.resource,
+            customIcon = customIcon.resource,
+            themeOverride = themeOverride,
+            shouldAnimateOnAttached = shouldAnimateOnAttached
+        )
+    }
```
pmartinbTEF, line 91: could be moved to else block of line 119?

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
```diff
@@ -60,37 +108,183 @@ fun Feedbacks(
...
-                Checkbox(checked = showFirstButtonLoading, onCheckedChange = { showFirstButtonLoading = it })
-            }
-            Row {
-                Text("Info")
-                Checkbox(checked = type == FeedbackScreenView.TYPE_INFO, onCheckedChange = { type = FeedbackScreenView.TYPE_INFO })
-                Text("Success")
-                Checkbox(checked = type == FeedbackScreenView.TYPE_SUCCESS, onCheckedChange = { type = FeedbackScreenView.TYPE_SUCCESS })
-                Text("Error")
-                Checkbox(checked = type == FeedbackScreenView.TYPE_ERROR, onCheckedChange = { type = FeedbackScreenView.TYPE_ERROR })
+        ConstraintLayout {
```
pmartinbTEF, line 120: If this is common code for compose/classic it should be outside an else block no?

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
jdelga: I don't see you point, the flag `showFeedbackClassic` launches a new activity, while `showFeedbackCompose` draws over the same view.

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
jdelga: It is common, however, it is drawn in the same view as the Feedback itself, so we want to draw only one of them. When the compose feedback is shown, the inputs are hidden and they are shown when we press back and `showFeedbackCompose` is false again

library/src/main/java/com/telefonica/mistica/feedback/screen/README.md
```diff
@@ -28,9 +28,9 @@ In order to configure it for the specific purpose, it supports both **attribute
 	<attr name="feedbackFirstButtonLoadingText" format="string"/>
     <attr name="feedbackSecondButtonText" format="string"/>
     <attr name="feedbackSecondButtonAsLink" format="boolean"/>
-
     <attr name="feedbackFirstButtonOnClick" format="string"/>
     <attr name="feedbackSecondButtonOnClick" format="string"/>
+	<attr name="shouldAnimateOnAttached" format="boolean" />
```
pmartinbTEF, line 33: Fields information entry should be added for shouldAnimateOnAttached

pmartinbTEF (changes requested):
Great job! just added a few questions

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/activity/FeedbackScreenCatalogActivity.kt
jdelga: Keep in mind that this is the catalog, not the library, so if we have enabled `shouldAnimateOnAttached` then we don't need to perform the animation manually.

We only need it if we disable `shouldAnimateOnAttached`

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
pmartinbTEF: I would add a third var showCatalogInputs or something similar to avoid using showFeedbackCompose for two different things

jdelga: but "show catalog inputs" is the case when the two others is false.

library/src/main/java/com/telefonica/mistica/feedback/screen/README.md
jdelga: Done

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/CommonButtonLaunchBlock.kt
```diff
@@ -0,0 +1,43 @@
...
+    onLaunchClassic: () -> Unit,
+    onLaunchCompose: () -> Unit,
+){
+    Column(
+        modifier = modifier
+            .background(Color.White)
+            .fillMaxWidth(),
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        Spacer(modifier = Modifier.height(20.dp))
```
DevPabloGarcia, line 29: Question: Instead of using different spacers, we can set Row composable: modifier.padding(20.dp) ?

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
DevPabloGarcia, line 120: I would extract this code for a new file like: FeedbacksForm.kt

DevPabloGarcia (commented):
Nice approach! Check little comments

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/CommonButtonLaunchBlock.kt
jdelga: No problem, I don't see any difference between them

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Feedbacks.kt
jdelga: The feedback is going to be different for each component, I don't think it's reusable. However, I can move it to a different composable for cleanness, WDYT?

pmartinbTEF (approved):
Good job!

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/activity/FeedbackScreenCatalogActivity.kt
yceballost, line 1: To avoid bad usage in mistica catalog, can you change the custom icon to this
https://github.com/Telefonica/mistica-icons/blob/production/icons/telefonica/light/lifeguard-float-light.svg

an example
image

yceballost, line 1: Its a bit weird to mix feedback typologies (Success, error, info) with custom typology. In my mind was better to change between typologies and inside the informative type, you can change to a custom icon.

Custom icon is just for INFO Screen

yceballost (changes requested):

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/activity/FeedbackScreenCatalogActivity.kt
jdelga: There are no mixed typologies, we have added a new one called "Custom" that allows a custom icon or animation. However, the icon defaults to the INFO one if it is not set.

In the other hand, previous type INFO does not have any changes.

yceballost: This approach is far from how it is approached from the design side. When a designer uses this in Figma, he will use the `InfoFeedbackScreen` and change the icon of this component, not use a `CustomFeedbackScreen`, fyi.

I think it's important that the design and development approach should be aligned so that when I get designs to the development teams, things appear coherent when the screens are analyzed.

for example
image

jdelga: I understood that with "custom" there were no limitations on the icon chosen. Is this fixed to some specific icons ?

jdelga: Captura de pantalla 2023-05-30 a las 13 02 46

jdelga: Ok then, thanks

DevPabloGarcia (approved):

jeslat (approved):

yceballost (approved):

PR #250, opened 2023-05-29
Generate colors, radius and text weights from Mistica tokens

jeslat:
Create a GitHub Action to generate the mistica tokens (currently colors) using the json file that it's generated by UX team. 

I've created a new module `tokens` which is responsible for generating all the files needed. This new module is invoked by the GitHub Action.

As we have to support the XML and the Compose implementation of Mistica, the code is generating both types of files. I've done some changes in the current Mistica implementation to be able to support the code generation easily:

- In Compose:
  - I've moved the `BrandPaletteColor` object and the definition of the light and dark colors to a different file which is named `XXXBrandColorsAutogenerated`. With this, I only have to generate code related to colors and not other things like what's the compatibility theme.
- In XML:
  - I've created a new style per brand `MisticaTheme.XXXX_Customizations` to include there the things that we touch manually because they are not defined in the tokens files. This style is used as the parent of the autogenerated one (that uses the same name that previously) so it's retro-compatible.

I've used a couple of libraries to generate the XML files and the Kotlin files:
- https://github.com/redundent/kotlin-xml-builder
- https://github.com/square/kotlinpoet

You can see here the files that are generated by UX team: https://github.com/Telefonica/mistica-design/tree/production/tokens

### ☑️ Checks

- I've tested the GitHub Action manually with ACT.
- I've generated the files locally using the jsons from the mistica-design repository.


.github/workflows/import-design-tokens.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          repository: Telefonica/mistica-design
+          path: '.github/mistica-design'
+          ref: ${{github.event.inputs.ref}}
+          token: ${{ secrets.FLOW_GITHUB_TOKEN }}
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'liberica'
+          java-version: '18'
```
yamal-alm, line 26: isn't it enough with the java version included in ubuntu-latest?

library/src/main/java/com/telefonica/mistica/compose/card/highlightedcard/HighLightedCard.kt
```diff
@@ -10,7 +10,6 @@ import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.foundation.verticalScroll
```
yamal-alm, line 13: don't we have detekt configured for mistica?

.github/workflows/import-design-tokens.yml
jeslat: It is, I added it to test it locally and I forget to remove it

library/src/main/java/com/telefonica/mistica/compose/card/highlightedcard/HighLightedCard.kt
jeslat: No, we don't have it

library/src/main/res/values/colors_blau_autogenerated.xml
```diff
@@ -0,0 +1,157 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dagonco, line 1: Not a strong opinion but I'm fine using same file name as before without "autogenerated" suffix.

.github/workflows/import-design-tokens.yml
```diff
@@ -0,0 +1,39 @@
...
+          cache: 'gradle'
+
+      - name: Generate skins from design tokens
+        run: ./gradlew run
+
+      - name: Create Pull Request
```
yamal-alm, line 32: question (I could have missed somthing): isn't the mistica-design repo included in the commit?

library/src/main/res/values/colors_blau_autogenerated.xml
```diff
@@ -0,0 +1,157 @@
...
+<resources>
+	<!-- Code generated automatically, DO NOT EDIT manually. Use Import design tokens GitHub Action. -->
+	<color name="blau_color_blauBluePrimary">
+		#00B6F1
+	</color>
```
dagonco, line 6: That's an ugly format!
Just curious, why did you use a library instead java's XMLStreamWriter class?

.github/workflows/import-design-tokens.yml
```diff
@@ -0,0 +1,39 @@
...
+          token: ${{ secrets.FLOW_GITHUB_TOKEN }}
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'liberica'
+          java-version: '18'
+          cache: 'gradle'
+
+      - name: Generate skins from design tokens
+        run: ./gradlew run
```
yamal-alm, line 30: I'm not very familiar with this.. `run` executes the main() function from `TokensGeneration.kt` file right? Have you considered creating a new gradle task called `generateMisticaTokens` or something like that? In that way, we are more aware about what this action is doing

jdelga:
This looks great @jeslat !! Only one comment, and is that I don't think we need to name any file `autogenerated` or `customizations` the same way don't do it for the automated translations

library/src/main/java/com/telefonica/mistica/compose/card/highlightedcard/HighLightedCard.kt
yamal-alm: oki, thx. Then I think we should, let's open that discussion offline

.github/workflows/import-design-tokens.yml
jeslat: I think you are right 😅

yamal-alm: a new step right before the commit removing the directory should be enough :)

library/src/main/res/values/colors_blau_autogenerated.xml
jeslat: You should not open this file, it just works 😄 

I use the library because I think the DSL is more beautiful that the XMLStreamWriter class, in any case, I can change this format to be just in one line.

jeslat: Done

library/src/main/java/com/telefonica/mistica/compose/card/highlightedcard/HighLightedCard.kt
nimeacuerdo: already identified some time ago with https://jira.tid.es/browse/ANDROID-11179 from https://confluence.tid.es/display/CTO/Unify+tooling+and+processes+between+different+projects

UPGRADING.md
```diff
@@ -36,3 +36,26 @@ Note that if your project is implementing 'catalog-compose', please consider to
 ```groovy
 debugImplementation "com.telefonica:mistica-catalog:$mistica_version"
 ```
+
+## Upgrade to version 9.0.0
+
+### Some colors have been renamed and deleted
+
+In order to align the color names with the one defined by UX team, some colors have been renamed. Some examples:
+- `buttonSecondaryBackgroundSelectedInverse` -> `buttonSecondaryBackgroundInverseSelected`
```
nimeacuerdo, line 45: The fact that the order of Inverse was inverted is so p🤯etic

.github/workflows/import-design-tokens.yml
jeslat: Done, in any case I will test the action once is merged

nimeacuerdo (approved):
Awesome work!

I just miss some documentation about the expected workflow of this tool for the contributors of this repo, maybe also pointing out things like the comments about autogenerated code that people will find, etc

dagonco (approved):
Currasso! 🥒

tokens/build.gradle
```diff
@@ -0,0 +1,26 @@
+plugins {
+    id 'java-library'
+    id 'org.jetbrains.kotlin.jvm'
+    id 'application'
+}
+
+mainClassName = 'com.telefonica.mistica.tokens.TokensGeneratorKt'
```
yamal-alm, line 7: this line is no longer needed right?

yamal-alm (approved):
awesome 👏

jeslat:
> Awesome work!
> 
> I just miss some documentation about the expected workflow of this tool for the contributors of this repo, maybe also pointing out things like the comments about autogenerated code that people will find, etc

@nimeacuerdo I've updated the readme with some info

README.md
```diff
@@ -157,6 +157,14 @@ allprojects {
...
 - Update the version of mistica to the one the snapshot was released with, sync your project and the dependency should be resolved with the artifact including your local changes.
 
+## Design tokens
+We have several tokens that are defined by the UX team (colors, radius and text presets) in a [JSON format](https://github.com/Telefonica/mistica-design/tree/production/tokens). Using these files, we generate XML and Kotlin files that are used in the library.
+
+You should not modify manually the generated files. All the generated files have a comment at the beginning to identify them.
+
+### How to update the design tokens
+The expected workflow is that UX team runs a GitHub action in their repository, which will invoke our `Import design tokens` action. This action will upload the generated files and create a new PR with the changes.
+
```
nimeacuerdo, line 167: Will this PRs be differentiated somehow from other, human-created, PRs? how will we be aware of them, because there will be someone assigned by default or what? 🤔

jeslat: The title will be `Update design tokens` always, I'm going to add a body in the PR to add a bit more information. 

Regarding reviewers, I've added `Android` team as a reviewer so two of us will be assigned to the PR automatically.

jeslat: Yep, maybe is a good option. Done

tokens/src/main/java/com/telefonica/mistica/tokens/TokensGenerator.kt
```diff
@@ -0,0 +1,5 @@
+package com.telefonica.mistica.tokens
+
+fun main() {
+    GenerateTokens().invoke()
```
jeprubio, line 4: I find weird having an invoke function which is not an operator. The same for GenerateXMLFiles().invoke()

jeslat: Oh I forget the `operator` in that function, in any case, I'm not sure about what is more weird:
`GenerateTokens().invoke()`
vs
`GenerateTokens()()`

yamal-alm: I agree with Josep, maybe some `TokenGenerator` class with a `generate` method would be better

tokens/src/main/java/com/telefonica/mistica/tokens/common/GetColorsWithAlpha.kt
```diff
@@ -0,0 +1,58 @@
...
+        brand: String,
+    ): Pair<String, String>? {
+        if (color.value.contains("rgba(")) {
+            val alpha = GenerateTokens.ALPHA_REGEX.find(color.value)?.value?.toDouble()
+
+            val colorValue = tokens.global.palette[color.description]?.value
+            if (alpha != null && colorValue != null) {
+                val alphaHex = alpha.toHex()
+                val colorName = getColorNameWithAlpha(brand, color.description, alpha)
+                val colorWithAlpha = "#" + alphaHex + colorValue.removePrefix("#")
```
jeprubio, line 44: I would probably have used string template in here instead of concatenating strings:
```
val colorWithAlpha = "#${alphaHex}${colorValue.removePrefix("#")}"
```
I wonder if it's only my android studio that warns me against concatenating strings and preferring the more idiomatic way of string templates.

tokens/src/main/java/com/telefonica/mistica/tokens/common/GetColorsWithAlpha.kt
```diff
@@ -0,0 +1,58 @@
+package com.telefonica.mistica.tokens.common
+
+import com.telefonica.mistica.tokens.GenerateTokens
+import com.telefonica.mistica.tokens.dto.ColorDTO
+import com.telefonica.mistica.tokens.dto.TokensDTO
+import java.util.Locale
+import kotlin.math.roundToInt
+
+class GetColorsWithAlpha {
```
jeprubio, line 9: Are there tests with some GetColorsWithAlpha invokations? Shouldn't be difficult to add few, I think.

tokens/src/main/java/com/telefonica/mistica/tokens/xml/GenerateAttributesFile.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.telefonica.mistica.tokens.xml.GenerateXMLFiles.Companion.VALUES_DIR
+import com.telefonica.mistica.tokens.xml.GenerateXMLFiles.Companion.capitalizeString
+import com.telefonica.mistica.tokens.dto.TokensDTO
+import org.redundent.kotlin.xml.XmlVersion
+import org.redundent.kotlin.xml.xml
+import java.io.File
+
+class GenerateAttributesFile {
+
+    operator fun invoke(jsonAdapter: JsonAdapter<TokensDTO>) {
```
jeprubio, line 16: I like the way you have broken the methods into small chunks, this one seems still a bit to big, could it be splitted in smaller private methods?

tokens/src/main/java/com/telefonica/mistica/tokens/TokensGenerator.kt
jeslat: Ok, understood. I've done some changes:
- `TokensGenerator` file has been renamed to `Main`
- `GenerateTokens` class has been renamed to `TokensGenerator` with a function called `generate`
- I've moved the dependencies of the rest of the classes to the constructor, instead of creating a new instance of them when it's necessary. With this change in the rest of the places, I can use the `invoke` without strange things.

tokens/src/main/java/com/telefonica/mistica/tokens/common/GetColorsWithAlpha.kt
jeslat: Agree, I don't know why my android studio doesn't warn me

tokens/src/main/java/com/telefonica/mistica/tokens/xml/GenerateAttributesFile.kt
jeslat: Agree, I've split into three more methods

tokens/src/main/java/com/telefonica/mistica/tokens/common/GetColorsWithAlpha.kt
jeslat: Let me try

jeslat: Done

tokens/src/main/java/com/telefonica/mistica/tokens/compose/GenerateBrandColors.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private fun getColorsConstructor(
+        brand: String,
+        colors: Map<String, ColorDTO>,
+        paletteClassName: String,
+    ): String {
+
+        var colorsConstructor = "${GenerateComposeFiles.MISTICA_COLORS}("
+
+        colors.forEach { color ->
```
haynlo, line 63: Consider to use destructuring declaration in that cases in order to avoid misunderstanding between color item and ColorDTO object and make it more readable.
```suggestion
        colors.forEach { (key, colorDto) ->
```
E.g. changing "color.value.value" to "colorDto.value"

haynlo (approved):
Awesome job!! 👏🏼

haynlo (approved):
Awesome job!! 👏🏼

.github/workflows/import-design-tokens.yml
```diff
@@ -0,0 +1,37 @@
...
+  import-design-tokens:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Checkout mistica-design repo with the tokens
+        uses: actions/checkout@v3
+        with:
+          repository: Telefonica/mistica-design
+          path: '.github/mistica-design'
```
idenjoe, line 19: Using the .github directory to checkout the mistica-design repository can result in a [git error](https://github.com/Telefonica/mistica-ios/actions/runs/5089898958/jobs/9149517214) by having it cached as a git submodule. I would use a temporary directory that is deleted from the runner at the end of all the jobs.

PR #259, opened 2023-06-06
ANDROID-13376 - Init FeedbackScreenView buttons at init

pmartinbTEF:
🎟️ Jira tickets

[Task](https://jira.tid.es/browse/ANDROID-13376)

Init FeedbackScreenView buttons at init to prevent [UninitializedPropertyAccessException](https://jira.tid.es/browse/ANDROID-13363)

Just init  buttons at init like the rest of UI elements

### ☑️ Checks

No tests needed


library/src/main/java/com/telefonica/mistica/feedback/screen/view/FeedbackScreenView.kt
```diff
@@ -228,6 +228,8 @@ class FeedbackScreenView : ConstraintLayout {
         errorReference = findViewById(R.id.error_reference)
         customContentContainer = findViewById(R.id.custom_content)
         buttonsContainer = findViewById(R.id.buttons_container)
+        firstButton = findViewById(R.id.first_button)
+        secondButton = findViewById(if (secondButtonAsLink) R.id.link_button else R.id.second_button)
```
jdelga, line 232: Cool! This is the place they belong 😄 

Can they be removed from `configureButtons` ?

pmartinbTEF: configureButtons is called from other functions (secondButtonAsLink could have changed for example). Anyway configureButtons remove all views from the buttons container so there is no problem in inflating the layout again

yamal-alm (approved):
looks good, thanks Pablo 👍

PR #260, opened 2023-06-07
ANDROID-13323 Create new vivo skin 

jeslat:
Create new vivo skin. Most of the code is autogenerated, I've only added the new skin in some places and modified the tokens generator module to support a brand with a two words name.

### ☑️ Checks
- [x] Tested with dark mode.

- You can review the generated catalog to check the new skin.


library/src/main/res/values/themes_vivoNew.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+	<style name="MisticaTheme.VivoNew_Customizations" parent="MisticaTheme.VivoNew_Base">
```
jeslat, line 4: Copy/paste from Vivo theme

tokens/src/main/java/com/telefonica/mistica/tokens/Brand.kt
```diff
@@ -0,0 +1,6 @@
+package com.telefonica.mistica.tokens
+
+data class Brand(
```
jeslat, line 3: I've created a new class to be able to have brands with a name that is different from the file name.

tokens/src/main/java/com/telefonica/mistica/tokens/TokensGenerator.kt
```diff
@@ -23,7 +23,14 @@ class TokensGenerator(
     }
 
     companion object {
-        val BRANDS = listOf("blau", "movistar", "o2", "telefonica", "vivo")
+        val BRANDS = listOf(
```
jeslat, line 26: This is the main change in tokens generator, the rest most of them are just renamings.

DevPabloGarcia (approved):
Nice Job!

yceballost (approved):

catalog/src/main/java/com/telefonica/mistica/catalog/ui/CatalogMainActivity.kt
```diff
@@ -42,6 +47,7 @@ class CatalogMainActivity : AppCompatActivity() {
             ComponentStyle("Movistar", R.style.MisticaTheme_Movistar, MOVISTAR),
             ComponentStyle("O2", R.style.MisticaTheme_O2, O2),
             ComponentStyle("Vivo", R.style.MisticaTheme_Vivo, VIVO),
+            ComponentStyle("Vivo New", R.style.MisticaTheme_VivoNew, VIVO_NEW),
```
gmerinojimenez, line 50: is this the official name? in case it isn't, what about trying to avoid the word "new"? Vivo v2, or something like that?

gmerinojimenez (approved):

catalog/src/main/java/com/telefonica/mistica/catalog/ui/CatalogMainActivity.kt
jeslat: Yes, it's the name that we have agreed on between the design team and the rest of the platforms. If we have a new vivo skin in the meanwhile, it will be `NewVivoNew` 😝

PR #262, opened 2023-06-12
ANDROID-12018 Add new text presets and text weights tokens 

jeslat:
In the new Vivo skin, we will need to use a new font, this font doesn't have `medium` weight so we need to be able to customize the text weight per brand in some components.

You can see here the text weight token definition: https://www.figma.com/file/PApssATUkePZnIOB4GmAeI/branch/XrVdts3xEHrQAFC3ztaQgE/%F0%9F%94%B9-%E2%80%8E%E2%80%8ETokens-definition?type=design&node-id=2190-1642&t=lPQuMjyAKzCyLWGk-0

In order to achieve this, I've created new text presets that inherit from the previous text preset that these components were using but override the text weight.

There are three commits where you can see each step:
* In the first one, I'm adding the tokens (many changes are autogenerated). I've included a small change in tokens generator to generate a different name for the text style attributes due to some collision.
* In the second one, I'm adding to the catalog the compose implementation of the badge component. It's not related to this PR but I've found that it was missing while I was testing this.
* In the third one, I'm applying the new text presets to the components.

### ☑️ Checks

- Check the generated catalog


jdelga (approved):
Cool!

jeslat:
TODO: Migrate Link text in Title component

gmerinojimenez (approved):
Nice job!

PR #268, opened 2023-06-20
ANDROID-13440 Fix dark mode tokens with alpha

haynlo:
Fix dark mode tokens with alpha

Add alpha check in dark theme files generation method.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

Evidence: https://github.com/Telefonica/mistica-android/pull/269 (Updated)

### DIFF
--- a/tokens/src/main/java/com/telefonica/mistica/tokens/xml/GenerateXMLFiles.kt
+++ b/tokens/src/main/java/com/telefonica/mistica/tokens/xml/GenerateXMLFiles.kt
@@ -13,6 +13,7 @@ import com.telefonica.mistica.tokens.TokensGenerator.Companion.REGULAR
 import com.telefonica.mistica.tokens.common.GetBorderRadiusName
 import com.telefonica.mistica.tokens.common.GetColorNameWithAlpha
 import com.telefonica.mistica.tokens.common.GetColorsWithAlpha
+import com.telefonica.mistica.tokens.dto.ColorDTO
 import com.telefonica.mistica.tokens.dto.TokensDTO
 import org.redundent.kotlin.xml.Node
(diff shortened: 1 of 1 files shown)


tokens/src/main/java/com/telefonica/mistica/tokens/xml/GenerateXMLFiles.kt
```diff
@@ -180,9 +180,20 @@ class GenerateXMLFiles(
...
+                                -"@color/$colorValue"
+                            }
+                        }
+                    } else {
+                        "item" {
+                            attribute("name", colorName)
+                            -"@color/${brandName}_color_${color.value.description}"
+                        }
                     }
                 }
```
jeslat, line 198: I would extract all this code to a new method to reuse it in light and dark method:
```
    private fun Node.colorItem(color: Map.Entry<String, ColorDTO>, brandName: String) {
        val colorName = if (color.key == "controlActivated") {
            "colorControlActive"
        } else {
            "color${color.key.capitalizeString()}"
        }

        if (color.value.value.contains("rgba(")) {
            val alpha = ALPHA_REGEX.find(color.value.value)?.value?.toDouble()
            if (alpha != null) {
                val colorValue = getColorNameWithAlpha(brandName, color.value.description, alpha)
                "item" {
                    attribute("name", colorName)
                    -"@color/$colorValue"
                }
            }
        } else {
            "item" {
                attribute("name", colorName)
                -"@color/${brandName}_color_${color.value.description}"
            }
        }
    }
```

jeslat (approved):
good job!

pmartinbTEF (approved):
LGTM

PR #272, opened 2023-06-23
ANDROID-13505 Add assignees to update design tokens PRs

yamal-alm:
Add workflow trigger author as assignee to "update design tokens" PRs

Use `assignees` field from the action we are already using to create the PR (https://github.com/peter-evans/create-pull-request)

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.


jeslat (approved):

haynlo (approved):
👍🏼

PR #274, opened 2023-06-29
ANDROID-13458 - Test tag in list row icon

DevPabloGarcia:
Add test tag to List Row Icon

Use this PR to deprecate current function and create new one to only pass icon res and flag to set circle.

* Create new Composable function as a copy of old ListRowItem function
* Remove icon composable param
* Add iconResId and show icon in circle params
* Replace catalog with new composable function

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [ ] 🖼️ Screenshots/Videos
- [ ] ...


library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -42,6 +43,50 @@ import com.telefonica.mistica.compose.theme.brand.MovistarBrand
 
 @ExperimentalMaterialApi
 @Composable
+@Suppress("DEPRECATION")
```
jeprubio, line 46: 🤔  I guess this `@Supress("DEPRECATION")` is because it's using the other deprecated one inside, could we instead extract a private composable that is common and use it in both public methods so this suppress is not needed? Perhaps rename the current ListRowItem to something else (ListRowItemImpl ?) and make it private and then create the public deprecated one that just invokes that. It's an idea.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -115,7 +160,9 @@ fun ListRowItem(
             modifier = rowModifier.height(IntrinsicSize.Min)
         ) {
             if (icon != null) {
-                icon()
+                Box(modifier = Modifier.testTag(ListRowItemTestTags.LIST_ROW_ITEM_ICON)) {
```
jeprubio, line 163: Should this also have testTagAsResourceId so that it's also available from UiAutomator and other testing frameworks that have no access to test tag?

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -184,6 +231,13 @@ fun ListRowItem(
     }
 }
 
+@Composable
+private fun ListRowIcon(@DrawableRes iconResId: Int?) {
+    if (iconResId != null) {
+        Image(painter = painterResource(id = iconResId), contentDescription = null)
```
jeprubio, line 237: Are you sure this contentDescription has to be null? Is this icon not useful at all for accessibility?

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -42,6 +43,50 @@ import com.telefonica.mistica.compose.theme.brand.MovistarBrand
 
 @ExperimentalMaterialApi
 @Composable
+@Suppress("DEPRECATION")
+fun ListRowItem(
+    modifier: Modifier = Modifier,
+    @DrawableRes iconResId: Int? = null,
```
jeprubio, line 49: Another approach could be using the "old version" and passing the composable but as a sealed class/interface so not any composable can be passed. I wonder if it's worth trying.

jeprubio (commented):
Good work, few comments.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
DevPabloGarcia: are we using testTagAsResourceId in other implementation? I think we aren't

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
DevPabloGarcia: Done!

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
DevPabloGarcia: As suggest @yamal-alm I will create an impl function, WDYT?

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -42,6 +43,50 @@ import com.telefonica.mistica.compose.theme.brand.MovistarBrand
 
 @ExperimentalMaterialApi
 @Composable
+@Suppress("DEPRECATION")
+fun ListRowItem(
```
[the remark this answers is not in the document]
DevPabloGarcia: Nice approach! changed :)

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
jeprubio: I think this answer is for the other comment, that would solve the first of my comments but not this one

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
DevPabloGarcia: Done!

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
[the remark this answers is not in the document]
DevPabloGarcia: Sure! done :)

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
yamal-alm: that's a good idea!

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
jeprubio: OK, no problem, this was a suggestion, I think it could be useful for QA but it was just an idea.

yamal-alm: I think you have a point there. However to this day, there hasn't been any problem with just `testTag`. But, this is something that you can bring up on next Weekly to discuss :)

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
jeprubio: We have had a call, it's not just a restricted hierarchy as it would be a sealed class (any icon drawable can be passed and it doesn't have to be restricted) so it can be done with a data class instead.

DevPabloGarcia: Done with data class

jeprubio (approved):
Nice job 👏

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -253,6 +281,25 @@ object ListRowItemTestTags {
     const val LIST_ROW_ITEM_TITLE = "list_row_item_title"
 }
 
+data class ListIconConfig(
+    @DrawableRes val iconResId: Int? = null,
+    val showIconCircle: Boolean = false,
```
yamal-alm, line 286: will be the circle background color configurable?

jeprubio: It is, it's using CompositionLocals:
```
@Composable
fun Circle(
    color: Color = MisticaTheme.colors.neutralLow,
    size: Dp = 40.dp,
    content: @Composable (() -> Unit),
)
```
If it should be by params I would consider adding those 3 in the data class as circleParams property instead of showIconCircle, but I don't know if it's needed for now or worth doing it now 🤷

yamal-alm: I think we should have a well defined param/interface right now so we avoid breaking changes in the future in case we want to extend the configuration options. WDYT about discussing it with the rest of the team. Maybe we don't always need a circle, or to provide a resId, etc etc

yamal-alm (approved):
thanks Pablo :)

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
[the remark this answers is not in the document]
DevPabloGarcia: As we talk, all changes will be in other PR: https://jira.tid.es/browse/ANDROID-13531

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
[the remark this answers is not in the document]
DevPabloGarcia: As we talk, all changes will be in other PR: https://jira.tid.es/browse/ANDROID-13531

jeprubio (approved):

PR #275, opened 2023-07-03
Cancel ongoing debug_catalog.yml jobs

yamal-alm:
Cancel ongoing debug_catalog.yml jobs

Add cancel-in-progress flag to debug_catalog.yml worflow and also avoid its execution when we have modified files that matches the following regex:
- `**.md`
- `doc/**`
- `tokens/**`
- `.github/**`

### ☑️ Checks
No checks needed

- Debug catalog action was not triggered when a file inside the discarded ones is modified


dagonco (approved):
🥒🥒

PR #276, opened 2023-07-05
Post catalog public URL with specific version

jeslat:
Now that AppCenter supports public URLs that point to a specific version, we don't need anymore the logic to obtain the private URL that expires.

- I've tested locally using ACT:
```
[Generate debug catalog app/debug-catalog]   ✅  Success - Main Get version URL
[Generate debug catalog app/debug-catalog]   ⚙  ::set-output:: appcenter_url=https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/259
```

### DIFF
--- a/.github/workflows/debug_catalog.yml
+++ b/.github/workflows/debug_catalog.yml
@@ -34,11 +34,7 @@ jobs:
 
       - name: Get version URL
         id: url-request
-        uses: fjogeleit/http-request-action@v1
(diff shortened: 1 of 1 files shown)


yamal-alm (approved):
niiice

DevPabloGarcia (approved):

PR #277, opened 2023-07-05
Android-13531 Update list row icon

DevPabloGarcia:
Update list row icon to don't let developers add any kind of views.
Add new list spec
* load url image
* asset image with aspect ratio 1:1
* asset image with aspect ratio 16:9
* asset image with aspect ratio 7:10

* Update compose views 
* * Create new sealed class with all image types
* * Create draw function foreach new image types
* Update xml views
* * Add new aspect ratio images to current implementation
* Update Catalog

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [ ] 🖼️ Screenshots/Videos
- [ ] ...

https://github.com/Telefonica/mistica-android/assets/42326032/7d0865d1-08a8-4c23-b3c5-aeb37e631fc1

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
@@ -1,10 +1,8 @@
 package com.telefonica.mistica.compose.list
 
-import androidx.annotation.DrawableRes
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.wrapContentSize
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/telefonica/mistica/compose/list/README.md
```diff
@@ -42,22 +42,24 @@ enum class BackgroundType {
 Following the order in the image, the types are: normal, boxed and boxed-inverse. Take into account that `normal` is full width.
```
jeslat, line 43: Update the XML readme too https://github.com/Telefonica/mistica-android/tree/main/library/src/main/java/com/telefonica/mistica/list

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -106,7 +106,10 @@ class ListRowView @JvmOverloads constructor(
     @IntDef(
         TYPE_IMAGE,
         TYPE_SMALL_ICON,
-        TYPE_LARGE_ICON
+        TYPE_LARGE_ICON,
+        TYPE_IMAGE_1_1,
+        TYPE_IMAGE_7_10,
+        TYPE_IMAGE_16_9
```
jeslat, line 112: You have to add these values to existing ones in `attrs_components.xml` to be able to set them from XML

library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    @Composable
+    private fun SmallAsset.DrawSmallAsset() {
+        val painter = when {
+            iconResId != null -> painterResource(iconResId)
+            iconUrl != null -> rememberAsyncImagePainter(
+                ImageRequest.Builder(LocalContext.current)
+                    .data(iconUrl)
+                    .build()
+            )
```
yamal-alm, line 111: I miss some exposed configuration to the user of this interface. Url images sometimes need a placeholder while they are being downloaded, an error asset when the download fails, a transition between the placeholder and the final image, etc. Maybe we should try to expose just a `Painter` and let the client app use whatever it wants no matter if the image comes from an url or a resource Id. What do you think? :)

yamal-alm (commented):
just a couple of comments

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
DevPabloGarcia: Done !

library/src/main/java/com/telefonica/mistica/compose/list/README.md
DevPabloGarcia: Done !

library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
DevPabloGarcia: Nice approach, I think it is the best solution, will change :)

jeslat (approved):
nice 👍

yamal-alm (approved):
nice job 👏

AnaMontes11 (approved):

PR #280, opened 2023-07-25
Android 13593 unify id button

DevPabloGarcia:

❗ About 20 files are only import changes ❗ 

Unify Button element Ids
- button
- button_text
- button_text_loading
- button_progress
- button_icon

We found a problem with xml element. We can reach icon and text subelement in order to set the new ids.
This way, we changed this element to wrap the compose id.

* Delete Button and ProgressButton element
* Wrap Compose Button in new XML Button
* Change all instances of old buttons/progress button to use new Button element

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [ ] 🖼️ Screenshots/Videos
- [ ] ...


UPGRADING.md
```diff
@@ -63,3 +63,20 @@ Other colors have been removed because UX team has decided to remove them from t
 ### Changes in FeedbackScreenView
 - `TYPE_CUSTOM` has been removed. Use `TYPE_INFO` instead, it allows custom icon or animation.
 - `shouldAnimateOnAttachedToWindow` is now private. Use method `setShouldAnimateOnAttached()` or xml attribute `app:shouldAnimateOnAttached` instead.
+
+## Upgrade to version 9.1.0
```
jdelga, line 67: Current version is 9.2.0, next should be at least 9.3.0

jdelga (commented):
This PR consists mostly in refactoring the usages of `Button` in catalog and removing `ProgressButton`, where I would expect the new common internal ids

Is this really affecting the test tags ?

PR #292, opened 2023-08-04
ANDROID-13479 buttonLink with chevron

jeprubio:
### :tickets: Jira ticket
[ANDROID-13479](https://jira.tid.es/browse/ANDROID-13479)

To add a chevron to buttons so that we can use and enable it on links when there is navigation.

- Adding the new chevron image which is 8x20 (there was an icn_arrow but that one was 24x24 and it couldn't be tinted when using the colorNeutralMedium).
- Using that chevron image from compose with it's 2dp padding and adding it to the catalog.
- Using that chevron image from traditional views and adding it to the catalog.
- That MisticaTheme in the preview of the buttons was to fix the previews as the previews of the buttons in Android Studio are failing when the corners are not defined.

### ☑️ Checks
- [x] Tested with dark mode.
- [ ] Tested with API 23.

It's visible in the compose catalog.
- [x] 🖼️ Screenshots/Videos
Classic views light
classic_light
Classic views light inverse
classic_light_inverse
Classic views dark
classic_dark
Classic views dark inverse
classic_dark_inverse
Compose light
compose_light
Compose light inverse
compose_light_inverse
Compose dark
compose_dark
Compose dark inverse
compose_dark_inverse


catalog/src/main/res/layout/screen_buttons_catalog.xml
```diff
@@ -418,10 +435,27 @@
...
-                android:layout_marginBottom="8dp"
                 android:enabled="false"
                 android:text="Link Disabled" />
 
+            <com.telefonica.mistica.button.Button
+                style="@style/AppTheme.Button.LinkButtonInverse"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="8dp"
+                android:text="Link With chevron"
```
haynlo, line 446: Very minor suggestion: use the same text you use for Compose

haynlo (approved):
Looks good, nice job!! 🙌🏼

catalog/src/main/res/layout/screen_buttons_catalog.xml
jeprubio: Good shout, now it's using the same text on both versions, thanks!

dagonco (approved):
That's really cool!

library/src/main/res/drawable/icn_chevron.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="8dp"
+    android:height="20dp"
```
aweell, line 3: The height of the icon should increase when the font size increases based on user preferences so it's always centered.

jeprubio: Where in the task is that specified?

aweell: > Where in the task is that specified?

The issue was reported yesterday by Vivo team https://github.com/Telefonica/mistica-design/discussions/1392 and discussed this morning in the Appscore | Design weekly. Since the development was ongoing I commented here. If it's better to create a separate ticket for this task just let me know and I'll do it.

> Would it still be an image scaled?

Yes it would be an image scaled, but, my mistake, the image should not appear deformed when doing so height and width should increase accordingly.

jeprubio: I don't know if it should be in another ticket, I did the work of that ticket last week, only did a rename this morning as web was using withChevron instead of hasChevron.
I don't see any problem in doing that in compose but I don't think it's possible to do that new requirements with an image inside a button on traditional views so this might be delayed indefinitely unless someone finds a solution, idk.

yceballost: just a pic of the current behavior with big sizes
image

(maybe we need to include the expected behavior in button specs)

jeprubio: Now it's scaling in compose, still not in traditional views but I've been told that must be done as well
chevron_normal_font_size_compose
chevron_large_font_size_compose

jeprubio: And I've created a custom font for the traditional views, could you recheck this when the catalog is recreated?

yceballost: Mmm I checked. The chevron is misaligned from the text.. that its weird because de asset height always has to fill the font line height.

Can you confirm to me if chevron height is covering always de height of the line height?
image

When you export svg, the svg has to be these dimensions
image

20px for height that fit with the 20px of line-height of the base font size. So if you increase font-size, the height of asset should increase proportionally

jeprubio: You are talking about compose or traditional views? In traditional views as I said the material buttons can't have a growing image so I did a trick by downloading the svg and then transforming it to a character in a custom font and then using that. Otherwise we should get rid of the material button and do a complete custom button ourselves (with it's different states, ripple, accessibility, focus...) which would be quite difficult to do it right. In compose the svg should have the same height the text has.

jeprubio: Now it's using a wrapper so that it uses compose under the hood also for classic views, could you recheck that?

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Buttons.kt
```diff
@@ -101,6 +101,24 @@ private fun Buttons(
...
                     isLoading = false,
                 )
+
+                if (it in listOf(ButtonStyle.LINK, ButtonStyle.LINK_INVERSE)) {
+                    CatalogButton(
+                        name = "Link With Chevron".lowercase().replaceFirstChar(Char::titlecase),
+                        style = it,
+                        withChevron = true,
+                        enabled = true,
+                        isLoading = false,
```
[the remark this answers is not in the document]
jeprubio: I agree that there are many parameters and we are adding one more "withChevron" but in order to remove some parameters I would consider using a data class instead. I don't think it's a good idea using a modifier for the button's state.

library/src/main/java/com/telefonica/mistica/button/MisticaButton.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import com.telefonica.mistica.R
+import com.telefonica.mistica.compose.button.Button
+import com.telefonica.mistica.compose.button.ButtonStyle
+import com.telefonica.mistica.compose.composeview.AbstractMisticaComposeView
+
+class MisticaButton @JvmOverloads constructor(
```
gmerinojimenez, line 14: What about `Button2`. I would avoid Mística prefix, since the rest of the components doesn't call MísticaXXX

library/src/main/java/com/telefonica/mistica/button/Button.kt
```diff
@@ -5,20 +5,14 @@ import android.util.AttributeSet
...
-    constructor(context: Context) : super(context)
-
-    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs)
-
-    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(
-        context,
-        attrs,
-        defStyleAttr
-    )
+class Button @JvmOverloads constructor(
```
gmerinojimenez, line 8: I would deprecate this one in favour of the new Button

library/src/main/java/com/telefonica/mistica/button/MisticaButton.kt
jeprubio: It was discussed on android devs, I first of all exposed the 2 solutions of using a different package or different name, I was told not to use a different package but a different name instead.

library/src/main/java/com/telefonica/mistica/button/Button.kt
jeprubio: I wouldn't yet, cause we are first testing this with links, if we do so it would add many warnings where it's not a link and detekt would fail in cases it shouldn't

UPGRADING.md
```diff
@@ -63,3 +63,10 @@ Other colors have been removed because UX team has decided to remove them from t
 ### Changes in FeedbackScreenView
 - `TYPE_CUSTOM` has been removed. Use `TYPE_INFO` instead, it allows custom icon or animation.
 - `shouldAnimateOnAttachedToWindow` is now private. Use method `setShouldAnimateOnAttached()` or xml attribute `app:shouldAnimateOnAttached` instead.
+
+## Upgrade to version 9.3.0
+MisticaButton is a wrapper to be used in traditional views that uses under the hood the Button() composable.
+We should start migrating the links from Button to MisticaButton when not using compose as the latest one adds the ability to set a chevron.
+Watch out that the new app:style attribute is needed to set the button style (with LINK / LINK_INVERSE for links).
+
+When this new compose wrapper is used on dialogs the ViewTreeLifecycle should be provided.
```
jeprubio, line 72: An example about this will be provided on next pr about the Sheet, which is waiting for this to be merged first.

aweell (approved):
LGTM

yceballost (approved):

library/src/main/java/com/telefonica/mistica/button/Button.kt
jeprubio: As discussed instead of deprecating I've added a comment, check if it's ok: [e986782](https://github.com/Telefonica/mistica-android/pull/292/commits/e9867829023bfbf365402a774e682e9232920dfc)
> // This should be deprecated soon, when it's confirmed no problem appears when using button2.Button instead

library/src/main/java/com/telefonica/mistica/button/MisticaButton.kt
jeprubio: Using button2.Button in the end.

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Buttons.kt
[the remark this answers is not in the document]
gmerinojimenez: in my opinion it's not a state, but the config it is used to draw it. The same way as many other modifier configurations

haynlo (approved):
Nice job!

dagonco (approved):
Cool!

PR #293, opened 2023-08-14
Feature/android 13547 sheet with action buttons

jeprubio:
### :tickets: Jira ticket
[ANDROID-13547](https://jira.tid.es/browse/ANDROID-13547)

Creating the Sheet Component Bottom Actions.

- Links with chevron is needed: https://github.com/Telefonica/mistica-android/pull/292
- Adding the new ComponentType: BottomActions
- Creating the new layout sheet_bottom_actions.xml just for the buttons.
- BottomActions.toView to display the new view with the buttons.
- Adding access to set the text and Chevron from the code in MisticaButton.
- Added withBottomActions() to be used by the bridge.

### ☑️ Checks
- [x] Tested with dark mode.
- [ ] Tested with API 23.

It can be accessed in the Sheet section by selecting Children Type: BottomActions.
- [x] 🖼️ Screenshots/Videos
BottomActions
https://github.com/Telefonica/mistica-android/assets/13270085/002c68b2-1f97-4f61-900f-5a1298c2902b


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         compose_core_version = '1.3.1'
         compose_compiler_version = '1.4.7'
         kotlin_version = "1.8.21"
-        androidx_appcompat_version = "1.5.1"
+        androidx_appcompat_version = "1.6.0"
```
jeprubio, line 8: This is unrelated to the task, I saw it could be updated and did it. No problem in reverting this if asked to.

library/build.gradle
```diff
@@ -84,6 +84,8 @@ dependencies {
 
     implementation "com.github.skydoves:balloon-compose:1.5.3"
 
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.6.1"
```
jeprubio, line 87: This lib is needed for the fix of the BottomSheetDialog in order to support compose in it.

library/src/main/java/com/telefonica/mistica/button2/Button.kt
```diff
@@ -17,7 +17,7 @@ class Button @JvmOverloads constructor(
     defStyleAttr: Int = 0,
 ) : AbstractMisticaComposeView(context, attrs, defStyleAttr) {
 
-    private var text: String by mutableStateOf("")
+    var text: String by mutableStateOf("")
```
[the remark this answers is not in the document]
jdelga: But in the case of the chevron you have kept it private and added a setter method `showChevron()`

I would do the same here.

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
```diff
@@ -204,6 +246,35 @@ private fun ListInformative.toView(context: Context): View =
...
+        .also { view ->
+            val primaryButton = view.findViewById<Button>(R.id.sheet_action_primary_button)
+            val secondaryButton = view.findViewById<Button>(R.id.sheet_action_secondary_button)
+            val linkButton = view.findViewById<com.telefonica.mistica.button2.Button>(R.id.sheet_action_link_button)
+            primaryButton.text = this.primaryButton.title
+            primaryButton.setOnClickListener {
+                onSheetTapped.onTapped(this.id, "PRIMARY")
+            }
+            secondaryButton.visibility = this.secondaryButton?.let { button ->
+                secondaryButton?.text = this.secondaryButton?.title.orEmpty()
```
jdelga, line 260: `this.linkButton` is already checked and it is available as non null as `button`

jdelga, line 260: `this.linkButton` is already checked and it is available as non null as `button`


```suggestion
                secondaryButton?.text = button.title
```

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
```diff
@@ -204,6 +246,35 @@ private fun ListInformative.toView(context: Context): View =
...
+                secondaryButton.setOnClickListener {
+                    onSheetTapped.onTapped(this.id, "SECONDARY")
+                }
+                View.VISIBLE
+            } ?: View.GONE
+            linkButton.visibility = this.linkButton?.let { button ->
+                linkButton?.text = this.linkButton?.title.orEmpty()
```
jdelga, line 267: Same:

```suggestion
                linkButton?.text = button.title
```

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
jeprubio: Yup, that makes sense, updated.
Thanks!

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
jeprubio: Updated. Thanks

library/src/main/java/com/telefonica/mistica/button2/Button.kt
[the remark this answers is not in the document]
jdelga: My suggestion was precisely having them private and use a setter

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
yceballost, line 1: Everything works fine! 

But the style of some elements still needs to be changed, as we added in the jira ticket:
We also took the opportunity to include a small restyling in this component (for better visualization you can see these updates marked in red in the specifications within anatomy and text) but they are the following:
-Title -> text-preset-5
-Subtitle -> text-preset-3
-Draggable area height -> 32px
-The padding between the text content and the interactive content becomes 8px.
-Action height in Action list type -> 72px

yceballost (changes requested):

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
jeprubio: If I change this it will be changed for the other Sheet types that are not action buttons, is that ok?

yceballost: These visual changes apply for all sheets :)

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
```diff
@@ -41,6 +54,24 @@ open class SheetView(
...
+        val activity = extractActivity(context)
+        val decorView = window?.decorView
+        decorView?.setViewTreeLifecycleOwner(activity as? LifecycleOwner)
+        decorView?.setViewTreeViewModelStoreOwner(activity as? ViewModelStoreOwner)
+        decorView?.setViewTreeSavedStateRegistryOwner(activity as? SavedStateRegistryOwner)
+    }
+
+    private fun extractActivity(context: Context?): Activity? {
+        var currentContext = context
+        while (currentContext != null) {
```
dagonco, line 67: Not sure if a loop is the best approach, just I'm afraid of infinite loops.
Could we have something more loop-proof and also, why is the reason to have this? Why actual view context is not enough?

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
```diff
@@ -204,6 +246,35 @@ private fun ListInformative.toView(context: Context): View =
         it.adapter = InformativeListAdapter(this.elements.mapToInformativeViewData())
     }
 
+private fun BottomActions.toView(context: Context, onSheetTapped: InternalOnSheetTapped, container: ViewGroup): View {
+    return LayoutInflater.from(context).inflate(R.layout.sheet_bottom_actions, container, false)
```
dagonco, line 250: IMHO, is kinda difficult to read this method. But It's just my point of view, not necessary to change if y'all fine.

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
jeprubio: It can't be infinite cause it's getting every time the baseContext, so at some point that must be null:
```
while (currentContext != null) {
    if (currentContext is Activity) return currentContext
    if (currentContext !is ContextWrapper) break
    currentContext = currentContext.baseContext
}
return null
```
It also only continues while that current context is not an Activity and is a ContextWrapper.
The code is from: https://stackoverflow.com/questions/72395779/using-a-composeview-inside-layout-or-a-bottomsheetdialog-throws-exception and it's because the BottomSheetDialog we have in the project is not implementing lifecycle so it crashes when using compose into it unless that lifecycle is implemented, which this is doing.

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
jeprubio: The signature follows the same way the other types of sheet follow, see: ListSingleSelection.toView, ListActions.toView, ListInformative.toView
I've extracted the code of the setBottomActionElements if that seems better [e847ee1](https://github.com/Telefonica/mistica-android/pull/293/commits/e847ee1c30102cfef55e9afa63307712bddd53fa)

library/src/main/java/com/telefonica/mistica/button2/Button.kt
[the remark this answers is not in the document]
jeprubio: Yup, in the Sheet.kt after the values are set I've tried adding:
```
GlobalScope.launch {
  while(true) {
      delay(1_000)
      linkButton?.withChevron = true
      delay(1_000)
      linkButton?.withChevron = false
  }
}
```
(sorry for the code, it's just for demonstration) and the chevron is updating all the time with the new values.

nimeacuerdo:
@jdelga @yceballost anything else from your side that blocks this PR?

jdelga (approved):

library/src/main/java/com/telefonica/mistica/sheet/Sheet.kt
yceballost: I can't see the last changes in the last update in Mística Catalog, can you check if the last version include this changes?

jeprubio: Checked. It does, indeed. The title is preset 5, the subtitle is preset 3 and the paddings are changed:

title_preset5

subtitle_preset3

PR #295, opened 2023-08-24
ANDROID-13681 Revert using JvmOverloads in Button

jeprubio:
Fix the problem that has appeared with the buttons that have no style associated.

Reverting the button.Button last change of using @JvmOverloads as that clearly breaks it.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

Before:
before
After:
after
- [ ] 🖼️ Screenshots/Videos
- [ ] ...


pmartinbTEF (approved):
LGTM

haynlo (approved):
👍🏼

PR #296, opened 2023-08-28
POC roborazzi screenshot tests

jeprubio:
POC Roborazzi tests

- Adding Roboelectric
- Adding Roborazzi following the README instructions.

### What is missing:
- Perhaps we want to store the screenshots inside the codebase instead of what they do for demonstration, we could check if passing the fileName is ok or if we have other problems then.
- Github integration, there is https://github.com/takahirom/roborazzi-compare-on-github-comment-sample we could play with it and check if that's ok or do we need something more.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

- [ ] 🖼️ Screenshots/Videos
- [ ] ...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,7 @@ buildscript {
 plugins {
     id 'org.jetbrains.kotlin.android' version '1.5.21' apply false
     id 'io.github.gradle-nexus.publish-plugin' version '1.1.0' apply false
+    id "io.github.takahirom.roborazzi" version "1.4.0" apply false
 }
 
 allprojects {
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -3,6 +3,7 @@ plugins {
     id 'org.jetbrains.kotlin.android'
     id 'kotlin-kapt'
     id 'maven-publish'
+    id 'io.github.takahirom.roborazzi'
 }
 
 android {
@@ -51,6 +52,12 @@ android {
         sarifReport true
         checkDependencies true
     }
+
+    testOptions {
+        unitTests {
+            includeAndroidResources = true
+        }
+    }
 }
 
 task sourceJar(type: Jar) {
@@ -87,6 +94,11 @@ dependencies {
     implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.6.1"
 
     testImplementation 'junit:junit:4.13.2'
+    testImplementation 'org.robolectric:robolectric:4.10.3'
+    testImplementation 'androidx.compose.ui:ui-test-junit4:1.0.5'
+    testImplementation 'androidx.compose.ui:ui-test-manifest:1.0.5'
+    testImplementation 'io.github.takahirom.roborazzi:roborazzi:1.4.0'
+    testImplementation 'io.github.takahirom.roborazzi:roborazzi-compose:1.4.0'
     testImplementation "androidx.compose.ui:ui-test-junit4:$compose_ui_version"
     testImplementation "org.mockito.kotlin:mockito-kotlin:4.0.0"
 
--- a/library/src/test/java/com/telefonica/mistica/compose/input/PasswordInputKtTest.kt
+++ b/library/src/test/java/com/telefonica/mistica/compose/input/PasswordInputKtTest.kt
@@ -1,0 +1,96 @@
+package com.telefonica.mistica.compose.input
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performClick
+import com.github.takahirom.roborazzi.captureRoboImage
+import com.telefonica.mistica.compose.theme.MisticaTheme
+import com.telefonica.mistica.compose.theme.brand.MovistarBrand
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.GraphicsMode
+
+@RunWith(RobolectricTestRunner::class)
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
+internal class PasswordInputKtTest {
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    @Test
+    fun `check the password is initially not visible`() = test {
+        `when PasswordInput`()
+
+        `then screenshot is OK`()
+        `then the password is not visible`()
+    }
+
+    @Test
+    fun `check the password is visible after the visibility button is clicked`() = test {
+        `given PasswordInput`()
+
+        `when the visibility button is clicked`()
+
+        `then screenshot is OK`()
+        `then the password is visible`()
+    }
+
+    @Test
+    fun `check the password is not visible after toggling twice the visibility button`() = test {
+        `given PasswordInput`()
+
+        `when the visibility button is clicked`(times = 2)
+
+        `then screenshot is OK`()
+        `then the password is not visible`()
+    }
+
+    private fun TestScope.`given PasswordInput`() {
+        `when PasswordInput`()
+    }
+
+    private fun TestScope.`when PasswordInput`() {
+        composeTestRule.setContent {
+            MisticaTheme(brand = MovistarBrand) {
+                PasswordInput(
+                    value = textValue,
+                    onValueChange = onValueChanged,
+                    label = "label",
+                )
+            }
+        }
+    }
+
+    private fun `when the visibility button is clicked`(times: Int = 1) {
+        repeat(times) {
+            composeTestRule.onNodeWithTag(TextInputTestTags.PASSWORD_VISIBILITY_TOGGLE).performClick()
+        }
+    }
+
+    private fun TestScope.`then the password is not visible`() {
+        composeTestRule.onNodeWithText(textValue).assertDoesNotExist()
+    }
+
+    private fun TestScope.`then the password is visible`() {
+        composeTestRule.onNodeWithText(textValue).assertIsDisplayed()
+    }
+
+    private fun `then screenshot is OK`() {
+        composeTestRule.onRoot()
+            .captureRoboImage()
+    }
+
+    private fun test(block: TestScope.() -> Unit) {
+        TestScope().block()
+    }
+
+    private class TestScope {
+        val textValue = "textValue"
+        var valueChanged = false
+        val onValueChanged: (String) -> Unit = { valueChanged = true }
+    }
+}
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 #Fri May 19 13:07:06 CEST 2023
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


library/src/test/java/com/telefonica/mistica/input/TextInputTest.kt
```diff
@@ -0,0 +1,63 @@
...
+@RunWith(RobolectricTestRunner::class)
+internal class TextInputTest {
+    @get:Rule
+    val rule = activityScenarioRule<DummyActivity>()
+
+    @Before
+    @Throws(Exception::class)
+    fun setUp() {
+        ShadowLog.stream = System.out
+    }
```
jeprubio, line 31: An example of how to print logs using roboelectric

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.telefonica.mistica.compose.theme.brand.Brand
+
+object ScreenshotUtils {
+    fun getScreenshotName(brand: Brand? = null, extra: String = ""): String {
+        return """screenshots/
+            ${TestUtils.findRunningTestMethodName()}
+            ${brandName(brand, prepend = "_")}${extra}.png"""
+            .replace("\\s+".toRegex(), "")
+    }
```
jeprubio, line 11: This returns the whole file path where the screenshot is saved, it could be changed if we somehow want to create different directories for the screenshots.

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.experimental.runners.Enclosed
+import org.junit.runner.RunWith
+import org.robolectric.ParameterizedRobolectricTestRunner
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(Enclosed::class)
+internal class ButtonKtTest {
```
jeprubio, line 30: The two internal classes could be split in ButtonKtTest and ButtonBehaviourTest for example in order not to need this Enclosed. I wonder what's best. For now I've left the Enclosed as this way when you are on Buttons.kt you might expect to find all the tests in here but I see no problem with spliting it.

library/src/test/java/com/telefonica/mistica/compose/input/PasswordInputKtTest.kt
```diff
@@ -0,0 +1,98 @@
...
+internal class PasswordInputKtTest {
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    @Test
+    fun `check the password is initially not visible`() = test {
+        `when PasswordInput`()
+
+        `then screenshot is OK`()
+        `then the password is not visible`()
```
jeprubio, line 31: As previously discussed we could just leave the screenshot test in here but I think this way the test is more readable with the other method of when the password is visible and when it's not. But these tests could only contain the screenshot test if requested to.

library/src/test/java/com/telefonica/mistica/input/TextInputTest.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.telefonica.mistica.testutils.ScreenshotUtils
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.shadows.ShadowLog
+
+@RunWith(RobolectricTestRunner::class)
+internal class TextInputTest {
```
jeprubio, line 23: This is a "classic views" xml test

library/src/test/res/layout/screen_inputs_catalog.xml
```diff
@@ -0,0 +1,355 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jeprubio, line 1: This layout is copied from the catalog as it has many different usages of the TextInput in different situations and the lib doesn't have visibility to the catalog. It's copied in the test resources directory.

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
nimeacuerdo: +1 to splitting these suites into two different ones, the Button test suite and the parameterised test one

library/src/test/java/com/telefonica/mistica/compose/input/PasswordInputKtTest.kt
nimeacuerdo: +1 to removing this last assertion if the screenshot test implicitly tests that fact

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    @RunWith(ParameterizedRobolectricTestRunner::class)
+    @Config(qualifiers = RobolectricDeviceQualifiers.Pixel5)
```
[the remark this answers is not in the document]
jeprubio: It's not necessary but it returns us bigger images, @pmartinbTEF liked it more this way when we checked it as for example here in the buttons it was challenging to spot the rounded corners with the default config whereas this way the rounded corners were clear.

screenshots.md
```diff
@@ -0,0 +1,42 @@
...
+            .captureRoboImage(ScreenshotUtils.getScreenshotName())
+
+For traditional views run an activity or check `TextInputTest.kt` example on how to use a dummy activity to load them inside.
+
+## Check that changes in the screen are detected
+`Compare Screenshots` job will run in each pr and will run the `verifyRoborazziDebug` gradle task that checks the screenshots. In case this fails it will 
+generate a report and upload it to azure and/or github artifacts.
+
+## Access to screenshots test report
+At the bottom of the summary of the action that has failed you'll see that a report has been uploaded either to azure or github artifacts (or both).
```
nimeacuerdo, line 32: to be updated to remove github artifacts given they are not a valid option for now

library/build.gradle
```diff
@@ -88,7 +98,15 @@ dependencies {
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation "androidx.compose.ui:ui-test-junit4:$compose_ui_version"
+    testImplementation 'androidx.compose.ui:ui-test-manifest:1.0.5'
+    testImplementation 'org.robolectric:robolectric:4.10.3'
+    testImplementation 'io.github.takahirom.roborazzi:roborazzi:1.5.0'
+    testImplementation 'io.github.takahirom.roborazzi:roborazzi-compose:1.5.0'
```
nimeacuerdo, line 104: roborazzi version could be extracted to a property

nimeacuerdo:
Please, update the "Checks" section, checking anything already verified. Also edit "How can I test this" with whatever content is relevant there (including any pending test with TODO or similar)

.github/workflows/update_screenshot_baseline.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Run Roborazzi Record
+        run: 'bash ./gradlew clean recordRoborazziDebug'
+
+      - name: Check Git status
+        run: 'git status'
+        shell: bash
```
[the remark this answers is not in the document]
jeprubio: This was copied from https://github.com/Telefonica/niji-for-home-android/blob/master/.github/workflows/run_acceptance_tests.yml as we thought this was necessary for the azure upload

library/build.gradle
jeprubio: Extracted [593539f](https://github.com/Telefonica/mistica-android/pull/296/commits/593539f83d8cedeea3b8644ffc02b8ed7ca8c546)

screenshots.md
jeprubio: True, forgot to remove them in here. Done [7599217](https://github.com/Telefonica/mistica-android/pull/296/commits/75992177d9c0254b4e33db48692abf0785b37211)

library/src/test/java/com/telefonica/mistica/compose/input/PasswordInputKtTest.kt
jeprubio: on, removed [da6d038](https://github.com/Telefonica/mistica-android/pull/296/commits/da6d038fdaa0e1970e55e9bb7213a746a311e3ef)

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
jeprubio: Split done [8ef2191](https://github.com/Telefonica/mistica-android/pull/296/commits/8ef2191eaf74f27c72e64900fba4936b1d66767c)

.github/workflows/update_screenshot_baseline.yml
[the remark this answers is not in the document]
nimeacuerdo: it justs prints git status info to the console, I would get rid of this here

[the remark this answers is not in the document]
jeprubio: OK, yup, then it makes sense, removed [626f9dc](https://github.com/Telefonica/mistica-android/pull/296/commits/626f9dc4a9062ae0a81c97fa1fbf63ad63bab1e3)

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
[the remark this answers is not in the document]
nimeacuerdo: I would be more explicit about that, then. Because I guess we want everyone to use this in their tests for the sake of coherence in tests reports. Maybe a base class with this config that any other test class extends from as the default approach?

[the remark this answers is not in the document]
nimeacuerdo: I'm totally up for considering other options, but think of inheritance as just one tool in the toolbox with its own perks. In this case, it's important to keep a config value in one place to cut down on repeating code and, more importantly, to avoid ending up with tests using different settings.

.github/workflows/update_screenshot_baseline.yml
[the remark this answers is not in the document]
jeprubio: If that is removed the screenshots are not uploaded to azure and it prints:

> Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/mistica-android/mistica-android/.github/shared-actions/azure/upload-to-storage'. Did you forget to run actions/checkout before running your local action?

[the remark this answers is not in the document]
jeprubio: Once that is restored

Screenshot 2023-10-09 at 12 53 05

Restored in [1f4193e](https://github.com/Telefonica/mistica-android/pull/296/commits/1f4193ec6eeab26b9ce691cfa4cef8d570bbad06)

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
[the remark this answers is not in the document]
jeprubio: Using ScreenshotsTest base class:
[363798d](https://github.com/Telefonica/mistica-android/pull/296/commits/363798deeef2fa94083edf681716b937bd7532c3)
[7a12e7c](https://github.com/Telefonica/mistica-android/pull/296/commits/7a12e7c464ddce501650e592bb9261e14766059d)

library/src/test/java/com/telefonica/mistica/compose/button/ButtonBehaviourTest.kt
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+    }
+
+    private fun TestScope.`when the button is clicked`() {
+        composeTestRule.onNodeWithText(textValue).performClick()
+    }
+
+    private fun TestScope.`then the onClickListener has been invoked`() {
+        assertTrue(clicked)
```
yamal-alm, line 52: what do you think about using a mock instead?

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.telefonica.mistica.compose.theme.brand.TelefonicaBrand
+import com.telefonica.mistica.compose.theme.brand.VivoBrand
+import com.telefonica.mistica.testutils.ScreenshotsTest
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.ParameterizedRobolectricTestRunner
+
+@RunWith(ParameterizedRobolectricTestRunner::class)
+internal class ButtonKtTest(private val brand: Brand, private val darkTheme: Boolean): ScreenshotsTest() {
```
yamal-alm, line 23: `ButtonKt`?

.github/workflows/compare_screenshots.yml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          repository: Telefonica/github-actions
+          token: "${{ secrets.NOVUM_PRIVATE_REPOS }}"
+          path: .github/shared-actions
+
+      - name: Upload reports to azure
+        if: failure()
+        uses: ./.github/shared-actions/azure/upload-to-storage
+        with:
+          azure-account-name: ${{secrets.AZURE_ACCOUNT_NAME}}
```
pmartinbTEF, line 102: This will obfuscate the url of the screenshots report. Could it be clear text?

jeprubio: You are right, this is the cause we have urls in the reports with the *** as for example: https://***.blob.core.windows.net/ci-container-1696848766116/mistica-screenshots-tests-report.tar.gz
If this was in clear text that would not happen. However, my understanding is that we were asked from server core not to do that an use the secret in here, perhaps because this is a public repo.

library/src/test/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+		package="com.telefonica.mistica"
+		>
+
+	<uses-permission android:name="android.permission.VIBRATE" />
```
pmartinbTEF, line 5: Why is it needed?

jeprubio: The reason of having this is because I copied the manifest from the library without the tests and then just added the application and activity blocks in there. This way we are using the same manifest as the library in prod for the tests and just adding the activity in there.

yamal-alm (approved):
nice! just some minor comments

.github/workflows/compare_screenshots.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: 'bash ./gradlew verifyRoborazziDebug'
+
+      - id: generate-html
+        name: Generate Html Report
+        if: failure()
+        env:
+          BRANCH_NAME: companion_${{ github.event.workflow_run.head_branch }}
+        shell: bash
+        run: |
+          mkdir reports
```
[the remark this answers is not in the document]
jeprubio: Yup, we could, I thought about this, but then I thought as the roborazzi lib is adding it's own html reports we could use them once that version is stable and we update and then remove this script. So I thought it's not worth moving it if it has to be removed later on.

library/src/test/java/com/telefonica/mistica/compose/button/ButtonBehaviourTest.kt
jeprubio: That would work as well and it wouldn't be a big deal but I prefer not to use mock when it's not needed as in here and all the compose ui tests I've seen until now. As my understanding is that using a mock framework adds a little performance overhead. But I'm also not against using mocks in tests like this, I just prefer not using them if possible.

.github/workflows/compare_screenshots.yml
```diff
@@ -0,0 +1,106 @@
...
+  CompareScreenshots:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Verify Screenshots (roborazzi)
+        run: 'bash ./gradlew verifyRoborazziDebug'
```
gmerinojimenez, line 17: I suggest using https://github.com/marketplace/actions/gradle-build-action to improve the performance

You can see examples in https://github.com/Telefonica/niji-for-home-android/blob/master/.github/workflows/autotest.yml

.github/workflows/compare_screenshots.yml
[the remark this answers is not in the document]
gmerinojimenez: I would use a custom action for it. It takes 5 min to create it. See https://github.com/Telefonica/novum-platform-android/tree/main/.github/actions/boot-emulator as an example

build.gradle
```diff
@@ -27,6 +28,7 @@ buildscript {
 plugins {
     id 'org.jetbrains.kotlin.android' version '1.5.21' apply false
     id 'io.github.gradle-nexus.publish-plugin' version '1.1.0' apply false
+    id "io.github.takahirom.roborazzi" version '1.5.0' apply false
```
gmerinojimenez, line 31: just to be sure, the plugin and the library use different versions?

library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt
jeprubio: The ButtonKtTest is the name android studio gives automatically to this test when done from the IDE, I can rename it to ButtonTest which is what I understand that you mean but then when navigating from the IDE with `command + shift + T` it won't find it anymore and will propose to create a new tests instead... so I think it's better that we can find that the class has tests and move quickly with the IDE than changing the name, wdyt?

library/src/test/res/layout/screen_inputs_catalog.xml
gmerinojimenez: would it be easier if we run the tests in the catalog? I know it may be weird, but it could be very easy to cover all the catalog screens with a test

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotsTest.kt
```diff
@@ -0,0 +1,34 @@
...
+        extra: String = "",
+    ) {
+        node.captureRoboImage(screenshotName(brand, darkTheme, extra))
+    }
+
+    private fun screenshotName(brand: Brand?, darkTheme: Boolean, extra: String) = ScreenshotUtils.getScreenshotName(
+        brand = brand,
+        extra = if (darkTheme) "_dark$extra" else extra
+    )
+}
```
gmerinojimenez, line 34: totally personal preference, but since those methods doesn't require anything from the class I would move them to pure functions

.github/workflows/compare_screenshots.yml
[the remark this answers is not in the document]
yamal-alm: even if it is going to be deleted, it doesn't feel clean to have a html in this main action file. We can wrap it inside a separated script or custom action as Guille said, invoke it from this main file and add a comment that this html generator execution is temporary and will be removed as soon as Roborazzi support is stable.

library/screenshots/check_TextInput_from_layout_1.png
pmartinbTEF, line 1: Is it a good idea to group several elements in one layout? Instead of doing parameterized test of each UI element like the Button example? I think it should be unified

jeprubio: In my opinion it's good offering multiple options and multiple ways to do this. In here we can use the layout from catalog as it has the different inputs and just use that instead of having to recreate the different TextInput states one by one and checking them separately, which would work as well. I'm just leaving more options to the ones creating the tests. With this way we could have mistica tested quite quickly and then decide wether it's worth doing it one by one or what. Wdyt?

.github/workflows/compare_screenshots.yml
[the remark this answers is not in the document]
jeprubio: OK, done [7fb6fc6](https://github.com/Telefonica/mistica-android/pull/296/commits/7fb6fc679a1001cbb21f292a9ec09c44fe61c5be)

library/src/test/res/layout/screen_inputs_catalog.xml
jeprubio: That was an option, I thought it made more sense to test the lib in the lib instead rather than depending on the catalog but yup, we could do that, idk what's best.

.github/workflows/compare_screenshots.yml
jeprubio: Done [e905f27](https://github.com/Telefonica/mistica-android/pull/296/commits/e905f27de8a506d9ea8b849159424140a25cdd06)

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotsTest.kt
gmerinojimenez: in my opinion there are two things: on one side is the qualifier, that force us to inherit and in the other the base methods. I don't have a strong opinion, so do what you prefer, but I don't think it is needed to add methods to this class (and couple the child tests to it) VS avoiding the inheritance.

Actually, if you ask me I would completely remove the ScreenshotsTest base class and instead add the qualifier to each test, like we do with some other tests like the acceptance.

library/src/test/res/layout/screen_inputs_catalog.xml
gmerinojimenez: Yep, it makes more sense, but a lot of the tests will duplicate the layout code, the other option may save us time

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotsTest.kt
jeprubio: Agree, but then we'll be back to the initial code in which I was asked not to add the qualifier to each test: https://github.com/Telefonica/mistica-android/pull/296#discussion_r1340334980

.github/workflows/compare_screenshots.yml
nimeacuerdo: exactly, that's what we have secrets for

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotsTest.kt
nimeacuerdo: Exactly... I'm not a fan of inheritance per se but here I see the same benefits we got from using a base class for unit tests in https://github.com/Telefonica/android-messenger/blob/integration/test-commons/src/main/java/com/tuenti/messenger/test/UnitTest.java to avoid repeating code in tests over and over again (and in for roborazzi, also possible errors to be introduced if we forget to set the qualifier or use a different one)

gmerinojimenez (approved):
Nice job!

nimeacuerdo (approved):
Get the ball rolling!

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotsTest.kt
jeprubio: Yup, in the end we agreed on leaving this base class for screenshot tests to avoid repeating code at least for now unless at some point we find another way to set the default qualifier and then this can be updated.

library/src/test/res/layout/screen_inputs_catalog.xml
jeprubio: I've removed the tests of the layout from the pr, we might want to reintroduce them in the catalog, or not. But I'm leaving them out of the scope for this pr.

library/screenshots/check_TextInput_from_layout_1.png
jeprubio: I've removed the tests of the layout from the pr, we might want to reintroduce them in the catalog, or not. But I'm leaving them out of the scope for this pr.

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotsTest.kt
```diff
@@ -7,7 +7,7 @@ import com.github.takahirom.roborazzi.captureRoboImage
 import com.telefonica.mistica.compose.theme.brand.Brand
 import org.robolectric.annotation.Config
 
-@Config(qualifiers = RobolectricDeviceQualifiers.Pixel5)
+@Config(sdk = [33], qualifiers = RobolectricDeviceQualifiers.Pixel5)
```
nimeacuerdo, line 10: why is the sdk needed here? what happens if we do not specify it?

jeprubio: It's not really needed, but it seems it's good to specify the sdk in case you want to be able to create the baseline from a different machine than the ci as otherwise I've read it might happen than the images not match on the ci if this sdk is not specified and specially if the baseline is created on a different OS than the ci is run.

nimeacuerdo: ok, +100 to have anything that could be affected by differences in the CI vs local environment specified in the config in order to prevent unexpected errors not related to actual bugs.

library/src/test/java/com/telefonica/mistica/testutils/ScreenshotUtils.kt
haynlo: For now I think it's better to keep this file path to save them, the [suggestion](https://confluence.tid.es/display/CTO/Roborazzi+screenshot+testing#:~:text=Are%20all%20screenshots%20in%20the%20/screenshot%20folder%20or%20grouped%20by%20brand%3F%20Or%20by%20file%3F%20Or%20by%20component%3F) commented in Confluence is something we can see later with the usage

haynlo (approved):
Great job!! 🚀

PR #299, opened 2023-09-06
Add instructions to add a new type of token

jeslat:
Add instructions to add a new type of token


DevPabloGarcia (approved):
Thanks!

haynlo (approved):
Thank you!

PR #300, opened 2023-09-07
ANDROID-13311 Evolve Feedback Screen

jeslat:
Evolve Feedback Screen:
- Reduce side margins from 24px to 16px
- Decrease icon size from 64 to 48
- Custom success animation for Vivo_New
- Update cascading opacity effect for screen content

* First commit: remove duplicated resources (they were the same the common and the Telefonica style).
* Second commit: modify margins of the layout
* Third commit: update success animation for Vivo new
* Fourth commit: update the entrance animation

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- You can use the generated catalog to see all the different combinations due to in some cases, the icon/animation changes per brand.

- Video with the feedbacks in Vivo-New and the new animations:

https://github.com/Telefonica/mistica-android/assets/2582348/727a4901-6685-40ae-b948-03b965b33e8d


jeprubio (commented):
This seem changes in the classic views, should it be also done in compose? Or am I missing something?

library/src/main/java/com/telefonica/mistica/feedback/screen/view/FeedbackScreenView.kt
```diff
@@ -462,7 +477,9 @@ class FeedbackScreenView : ConstraintLayout {
         const val TYPE_INFO = 2
 
         const val TEXTS_ANIMATION_DURATION = 800L
-        const val TEXTS_ANIMATION_DELAY = 200L
+        const val TITLE_ANIMATION_DELAY = 600L
+        const val SUBTITLE_ANIMATION_DELAY = 900L
+        const val EXTRAS_ANIMATION_DELAY = 1200L
```
jeprubio, line 482: I prefer to write it with underscores as thousands separator: `const val EXTRAS_ANIMATION_DELAY = 1_200L`

jeslat:
> This seem changes in the classic views, should it be also done in compose? Or am I missing something?

@jeprubio the compose implementation is just a wrapper of the XML view so no need to change anything

library/src/main/res/values/themes_vivoNew.xml
```diff
@@ -15,7 +15,7 @@
 		<!-- Feedbacks -->
 		<item name="feedbackScreenSuccessWithGradient">false</item>
 		<item name="feedbackScreenSuccessInverse">true</item>
-		<item name="feedbackScreenSuccessAnimation">@raw/feedback_success_vivo</item>
```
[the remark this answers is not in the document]
jeslat: `vivo` and `vivo_new` are two different brands/themes so we have to keep both. In this case, we use `feedback_success_vivo` into `themes_vivo.xml`

jeprubio (approved):

library/src/main/java/com/telefonica/mistica/feedback/screen/view/FeedbackScreenView.kt
jeslat: I agree with adding the underscore if the number is bigger, with just 4 digits, it is a bit on the limit for me. Anyway, I will change it.

jdelga (approved):
LGTM

yamal-alm:
> 📱 New catalog for testing generated: [Download](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/324)

Hi @AnaMontes11 This new catalog includes the new lottie asset for Vivo New. How about now? 🙂

Captura de Pantalla 2023-09-21 a las 9 32 53

AnaMontes11 (approved):
Now looks nice!

PR #301, opened 2023-09-12
ANDROID-11784 Fix Carousel component legacy implementation

pmartinbTEF:
Provide Carousel implementation without the need of a composable so it can be used in projects with no Jetpack Compose.

Provide a list of Views to CarouselView instead of a composable
Then CarouselView use [interoperability libraries](https://developer.android.com/jetpack/compose/migrate/interoperability-apis/views-in-compose) to render the views into a composable
Now there is no need to set a CarouselState to the CarouselView and the CarouselPageIndicatorView. Just set the CarouselPageIndicatorView to the CarouselView and the rest is done internally

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [x] Import Mistica in a project without Jepack Compose
- [x] Use a Carousel like in the Catalog app example


catalog/src/main/res/layout/carousel_fragment_catalog.xml
```diff
@@ -1,23 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
pmartinbTEF, line 2: Just to see the carousel in small devices

library/src/main/java/com/telefonica/mistica/card/CardView.kt
```diff
@@ -205,7 +205,7 @@ abstract class CardView @JvmOverloads constructor(
         cardContentView.pretitleTextView.setTextAndVisibility(pretitle)
     }
 
-    protected fun setSubtitle(subtitle: String) {
+    fun setSubtitle(subtitle: String) {
```
pmartinbTEF, line 208: To be able to set subtitle programmatically

library/src/main/java/com/telefonica/mistica/carousel/CarouselView.kt
```diff
@@ -17,20 +22,28 @@ class CarouselView @JvmOverloads constructor(
     defStyleAttr,
 ) {
 
-    private lateinit var carouselState: CarouselState
+    private var carouselState: CarouselState = CarouselState(PagerState(0))
```
pmartinbTEF, line 25: Init CarouselState and pass it to the CarouselPageIndicatorView

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/CarouselFragment.kt
```diff
@@ -35,35 +27,36 @@ class CarouselFragment : Fragment() {
         return binding.root
     }
 
-    @OptIn(ExperimentalPagerApi::class)
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        val carouselState = CarouselState(PagerState(0))
-        val pages = 6
```
gmerinojimenez, line 42: I'm not sure about this change, now the state is hidden and can't be controlled outside.

For example now is impossible to start at the page 1, wouldn't it better to be able to set it up from outside?

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/CarouselFragment.kt
```diff
@@ -35,35 +27,36 @@ class CarouselFragment : Fragment() {
...
-            .fillMaxWidth(),
-        image = MediaCardImage.MediaCardImageResource(card_image_sample),
-        tag = Tag("HEADLINE").withStyle(TYPE_PROMO),
-        title = "Page ${page + 1} ",
-        subtitle = "(position ${page})",
-        description = "Description",
-        primaryButton = Action("Primary") {},
-        linkButton = Action("Link") {}
-    )
 }
```
gmerinojimenez, line 61: I understand the reasoning to not using a composable to draw the items, but in fact, the initial idea was to expose the composable so it is possible to draw the inner content with compose (much easier) even in apps that doesn't use it. :)

Maybe an intermediate approach could be to expose both ways to render the content. A composable and a view wrapped by a ComposeView?

library/src/main/java/com/telefonica/mistica/card/CardView.kt
gmerinojimenez: Hmm I think they were protected on purpose (because it's accessed by the child classes). If we're going to make them public I would consider moving the rest of the methods

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/CarouselFragment.kt
```diff
@@ -35,35 +27,36 @@ class CarouselFragment : Fragment() {
...
-            .setItemCount(pages)
+            .setContent(getMediaCardsForCaroussel())
+            .setPagerIndicatorView(binding.carouselPageIndicatorView)
+            .setItemCount(MEDIA_CARDS_CAROUSEL_SIZE)
+    }
 
-        binding.carouselPageIndicatorView
-            .setState(carouselState)
-            .setPageCount(pages)
+    private fun getMediaCardsForCaroussel(): List<MediaCardView> {
```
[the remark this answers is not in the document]
pmartinbTEF: Done!

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/CarouselFragment.kt
pmartinbTEF: Ok, I reverted this change

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/CarouselFragment.kt
```diff
@@ -35,35 +27,36 @@ class CarouselFragment : Fragment() {
...
+                setTitle("Page$i")
+                setSubtitle("(position ${i - 1})")
+                setDescription("Description")
+                setPrimaryButtonText("Primary")
+                setLinkButtonText("Link")
+                setPrimaryButtonOnClick { Toast.makeText(requireContext(), "primaryButton$i", Toast.LENGTH_SHORT).show() }
+                setLinkButtonOnClick { Toast.makeText(requireContext(), "linkButton$i", Toast.LENGTH_SHORT).show() }
+                setCardAdditionalContent(null)
+            })
+        }
```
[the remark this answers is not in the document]
pmartinbTEF: I've done some tests and the app crashes if the data is changed after the view is init (with both approaches). Maybe I'm missing something. Anyway the carouselState change is reverted.

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/CarouselFragment.kt
pmartinbTEF: Good idea. Done!

library/src/main/java/com/telefonica/mistica/card/CardView.kt
pmartinbTEF: Reverted. I added a custom function to set subtitle to MediaCardView so it has no side effects

library/src/main/java/com/telefonica/mistica/carousel/CarouselView.kt
```diff
@@ -18,17 +22,43 @@ class CarouselView @JvmOverloads constructor(
...
 
     fun setState(carouselState: CarouselState): CarouselView = this.apply {
         this.carouselState = carouselState
+        initPageIndicator()
     }
 
     fun setItemCount(itemCount: Int): CarouselView = this.apply {
         this.itemCount = itemCount
+        initPageIndicator()
+    }
```
jeprubio, line 44: Are we sure carouselState is going to be set before setItemCount is invoked? because in initPageIndicator() is using that lateinit var carouselState and I wonder if it's going to be always set at this point. Perhaps it could have a default value instead? Only in case that could happen.

pmartinbTEF: There is no need to reference CarouselPageIndicatorView in CarouselView so I deleted it

jeprubio (approved):
Nice job 👏

gmerinojimenez (approved):
Thank you for the fix!!

jeprubio (approved):
Nice fix for ic_mistica_logo_text 🎉  Thanks!

yamal-alm (approved):

PR #302, opened 2023-09-20
ANDROID-13761 Adding semantics to the popover

gmerinojimenez:
Ticket: https://jira.tid.es/browse/ANDROID-13761

Popovers in compose are made with a custom library called balloon. It renders a composeView inside some view hierarchy.

So it is needed to set the semantics at this point.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

image
image


DevPabloGarcia (approved):

jmanriquehiberus (approved):
Nice job!

PR #304, opened 2023-09-25
Android-13783 Improve Callout buttons margin

haynlo:
Improve Callout buttons margin. Check 'How can I test this?' section to see the main differences.

* Improve Margins definition for Callout compose Component
* Remove link buttons horizontal margin

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

|  Before   | After   | Touch target size | 
| ------------- |-------------|-------------|
|  actual_bad  | after_good | after_touch_area |

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
@@ -84,7 +84,7 @@ fun Button(
                 }
                 .height(size.height)
                 .applyWidth(originalWidth),
-            contentPadding = PaddingValues(horizontal = size.contentPadding, vertical = 0.dp),
+            contentPadding = PaddingValues(horizontal = size.horizontalPadding, vertical = 0.dp),
             onClick = onClickListener,
             enabled = enabled,
             colors = style.buttonColors,
--- a/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
@@ -50,7 +50,7 @@ import com.telefonica.mistica.compose.theme.brand.MovistarBrand
 
 private val iconSpacing = 10.dp
 private val easing = CubicBezierEasing(0.77f, 0f, 0.175f, 1f)
-private const val CHEVRON_ASPECT_RATIO = 8f/20f
+private const val CHEVRON_ASPECT_RATIO = 8f / 20f
 
 @Composable
 fun Button(
@@ -62,6 +62,7 @@ fun Button(
     enabled: Boolean = true,
     @DrawableRes icon: Int? = null,
     withChevron: Boolean = false,
+    invalidatePaddings: Boolean = false,
     onClickListener: () -> Unit,
 ) {
 
@@ -84,7 +85,7 @@ fun Button(
                 }
                 .height(size.height)
                 .applyWidth(originalWidth),
-            contentPadding = PaddingValues(horizontal = size.horizontalPadding, vertical = 0.dp),
+            contentPadding = PaddingValues(horizontal = if (invalidatePaddings) 0.dp else size.horizontalPadding, vertical = 0.dp),
             onClick = onClickListener,
             enabled = enabled,
             colors = style.buttonColors,
--- a/library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
@@ -24,10 +24,9 @@ internal fun ButtonStyle.getButtonSizeCompose(): ButtonSizeConfig =
         ButtonStyle.PRIMARY_SMALL,
         ButtonStyle.SECONDARY_SMALL_INVERSE,
         ButtonStyle.SECONDARY_SMALL,
-        -> getSmallButtonSizeConfig()
         ButtonStyle.LINK,
         ButtonStyle.LINK_INVERSE,
-        -> getSmallButtonSizeConfig().copy(horizontalPadding = 0.dp)
+        -> getSmallButtonSizeConfig()
         else -> getDefaultButtonSizeConfig()
     }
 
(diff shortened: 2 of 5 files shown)


library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
```diff
@@ -6,14 +6,14 @@ import androidx.compose.ui.unit.Dp
...
-internal class ButtonSizeConfig(
+internal data class ButtonSizeConfig(
     val textStyle: TextStyle,
     val height: Dp,
     val minWidth: Dp,
     val iconSize: Dp,
     val progressBarSize: Dp,
     val progressBarStroke: Dp,
-    val contentPadding: Dp,
+    val horizontalPadding: Dp,
```
haynlo, line 16: Change it to a more accurate name.

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
```diff
@@ -24,9 +24,10 @@ internal fun ButtonStyle.getButtonSizeCompose(): ButtonSizeConfig =
         ButtonStyle.PRIMARY_SMALL,
         ButtonStyle.SECONDARY_SMALL_INVERSE,
         ButtonStyle.SECONDARY_SMALL,
+        -> getSmallButtonSizeConfig()
         ButtonStyle.LINK,
         ButtonStyle.LINK_INVERSE,
-        -> getSmallButtonSizeConfig()
+        -> getSmallButtonSizeConfig().copy(horizontalPadding = 0.dp)
```
haynlo, line 30: This is one of the reasons why the link button was not fully displayed. 
We don't need to apply a horizontal padding because link buttons don't have a drawable background.

library/src/main/java/com/telefonica/mistica/compose/callout/Callout.kt
```diff
@@ -88,27 +89,40 @@ fun Callout(
...
                         }
                         CalloutButtonConfig.SECONDARY -> {
-                            SecondaryButton(secondaryButtonText, onSecondaryButtonClick)
+                            SecondaryButton(text = secondaryButtonText, onClick = onSecondaryButtonClick)
                         }
                         CalloutButtonConfig.PRIMARY_AND_SECONDARY -> {
-                            PrimaryButton(primaryButtonText, onPrimaryButtonClick)
-                            SecondaryButton(secondaryButtonText, onSecondaryButtonClick)
+                            PrimaryButton(
+                                modifier = Modifier.padding(end = 16.dp),
```
haynlo, line 110: Another reason is that we were applying an end padding to all the buttons in Callout (primaries and secondaries). 

Now we only apply this padding if there's another button to be displayed on the right, so the secondary button or secondary link button don't need to apply this padding and they can use this space to display de text completely.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -109,7 +109,7 @@ fun ListRowItem(
 
 @ExperimentalMaterialApi
 @Composable
-fun ListRowItemImp(
+private fun ListRowItemImp(
```
haynlo, line 112: Component implementations should be private. 
I will notify other people to check possible breaking changes with that (Latch, Smart-Wifi and Novum have been checked and don't have any).

library/src/main/java/com/telefonica/mistica/compose/callout/Callout.kt
```diff
@@ -129,37 +142,41 @@ fun Callout(
 
 @Composable
 private fun PrimaryButton(
+    modifier: Modifier = Modifier,
     text: String?,
-    onClick: (() -> Unit)? = null
+    onClick: (() -> Unit)? = null,
```
jeprubio, line 147: There are few places in this pr that the optional modifier is set as the first parameter, is this some kind of rule in this project or something? I'm used to set it as the first parameter which is optional but I prefer to set non optional parameters on top, like the Text Composable for example, I would expect to see the text parameter on top.

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
jeprubio: Is this for removing always the horizontal padding of links? Are we sure this is the expected behaviour?

library/src/main/java/com/telefonica/mistica/compose/callout/Callout.kt
haynlo: I don't think we have a rule here. I always set the optional Modifier first for convenience but I don't have a strong opinion about that; your rule (non-optional first) makes more sense :)

library/src/main/java/com/telefonica/mistica/compose/button/ButtonSizes.kt
haynlo: I should be without padding the first time it was defined, but now I'm thinking that this change can lead to some kind of breaking change... Some people can be using this Button link keeping in mind this padding :(

As you can see in [here](https://user-images.githubusercontent.com/51969539/270404286-b794c9f7-a09a-4df8-a82b-63f727d214a5.png), the link button doesn't need any extra padding. This is a wrong initial implementation, but you are right, maybe changing this to zero directly is a bit risky.. I'd better comment this in the Teams channel

jeprubio (approved):
Nice

dagonco (approved):
Nice @haynlo! 😇

library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -62,6 +62,7 @@ fun Button(
     enabled: Boolean = true,
     @DrawableRes icon: Int? = null,
     withChevron: Boolean = false,
+    invalidatePaddings: Boolean = false,
```
jeprubio, line 65: My understanding was that we were going to set a parameter for padding values with a default value and this way it would allow that value to be set from outside, but as that padding value depends on the type of the button that makes everything more messy. I guess this makes sense having that into account.

jeprubio (approved):
LGTM. Thanks!

PR #306, opened 2023-10-17
New Snackbar spec

yamal-alm:
New Snackbar definition. Main changes:
- New possibility to set an indefinite duration of the Snackbar
- Dismiss button
  - It will appear when the duration was set to indefinite and there is no action or if the action triggers some event different than dismiss.
  - It also can be forced to appear

The new dismiss button is included in the Material 3 definition but it is not supported yet in the Material Android library that we are using. Instead I had to overwrite the layout of Snackbar class and do some adaptations to include this new button and logic:
- Retrieve `Snackbar::snackbarLayout` that is fortunately a public method and remove all of its child views
- Inflate our new custom Snackbar layout including the dismiss button and make it visible according to the previously described logic
- Implement existing logic where the "action" should be placed in a different line if it exceeds a max width.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [x] [Mistica App download link](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/361)
- [x] 🖼️ Screenshots

| Action and 10 seconds | Dismiss forced | Indefinite duration, long action and action that closes | Indefinite duration, long action that doesn't close | Indefinite duration, short action that close | Indefinite duration, short action that doesn't close |
| --- | --- | --- | ---- | --- | --- | 
|10_seconds_action|dismiss_forced|long_action_action_close|long_Action_close|short_action_action_close|short_action_close|


library/src/main/res/layout/snackbar_custom_layout_merge.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
```
yamal-alm, line 2: file almost indentical to the original layout from Material library implementation

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
```diff
@@ -82,47 +89,84 @@ open class SnackbarBuilder(view: View?, text: String) {
...
+            areSticky() -> SnackbarLength.INDEFINITE
+            snackbarLength == null -> if (actionListener == null) SnackbarLength.SHORT else SnackbarLength.LONG
+            else -> snackbarLength
+        }.duration()
+
+        val snackbar = inflateCustomSnackbar(duration)
+
+        snackbar.getCustomLayout().setText(text)
+        snackbar.setCustomAction()
+        snackbar.showDismissActionIfNeeded(hasInfiniteDuration = snackbarLength == SnackbarLength.INDEFINITE)
```
yamal-alm, line 103: all these customizations are managed inside `CustomSnackbarLayout` class

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
```diff
@@ -82,47 +89,84 @@ open class SnackbarBuilder(view: View?, text: String) {
...
+
+    private fun Snackbar.showDismissActionIfNeeded(hasInfiniteDuration: Boolean) {
+        val hasNoAction = actionText == null
+        val userCanNotDismissSnackbarWithoutTriggeringAnyAction = actionText != null && actionListener != null
+        val userShouldBeAbleToDismissSnackbar = hasInfiniteDuration &&
+                (hasNoAction || userCanNotDismissSnackbarWithoutTriggeringAnyAction)
+
+        if (isDismissible || userShouldBeAbleToDismissSnackbar) {
+            getCustomLayout().setOnDismissClickListener { dismiss() }
+        }
```
yamal-alm, line 150: if the snackbar has an indefinite duration, the user should be able to manually dismiss it. In case the action opens some screen or trigger some event that the user doesn't want to execute, then we should provide an alternative way to dismiss the snackbar by showing the dismiss button

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
```diff
@@ -52,7 +59,7 @@ open class SnackbarBuilder(view: View?, text: String) {
     }
 
     @JvmOverloads
-    open fun showCritical(snackbarLength: SnackbarLength = SnackbarLength.SHORT): Snackbar {
+    open fun showCritical(snackbarLength: SnackbarLength? = null): Snackbar {
```
jdelga, line 62: Why is the default changed ?

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
```diff
@@ -82,47 +89,84 @@ open class SnackbarBuilder(view: View?, text: String) {
...
-        val snackbar = Snackbar.make(view, text, duration)
-        setTextStyles(snackbar)
-        if (actionText != null) {
-            snackbar.setAction(actionText, actionListener)
-        }
-        if (callback != null) {
-            snackbar.setCallback(callback)
-        }
+            areSticky() -> SnackbarLength.INDEFINITE
+            snackbarLength == null -> if (actionListener == null) SnackbarLength.SHORT else SnackbarLength.LONG
```
jdelga, line 95: I don't understand the relation between the length and if the listener exists or not. IMHO it should default to the previous one which was `SHORT`

jdelga (commented):
Good job! I have added some questions.

jdelga:
What is the logic to change the snackbar color ?

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
yamal-alm: Previous implementation:
- There were only two possible snackbar lengths: "short" and "long".
- Even if "short" was set on purpose by the user or by default, if any action was provided, then the duration would be set to "long". This is like that by definition. 

New implementation:
- One more length choice: "indefinite".
- If no duration is explicitly selected, we use the same approach as before: if no action is set, then we use "short", if an action is provided then "long" is used instead.


However, I think `if (actionListener == null)` should be replaced with `if (actionText == null)`

yamal-alm:
> What is the logic to change the snackbar color ?

There is no logic, just the way the Snackbar is shown. We have two methods: `showCritical[ and `showInformative`

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
jdelga: Then it is more accurate now, thank you very much!

jdelga (approved):
Good job!

pmartinbTEF (approved):
Great job!

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
```diff
@@ -82,47 +89,84 @@ open class SnackbarBuilder(view: View?, text: String) {
...
-        setTextStyles(snackbar)
-        if (actionText != null) {
-            snackbar.setAction(actionText, actionListener)
-        }
-        if (callback != null) {
-            snackbar.setCallback(callback)
-        }
+            areSticky() -> SnackbarLength.INDEFINITE
+            snackbarLength == null -> if (actionText == null) SnackbarLength.SHORT else SnackbarLength.LONG
+            else -> snackbarLength
```
idenjoe, line 96: This code allows two cases that I think were not specified in the [specs](https://www.figma.com/file/yCHLIfy4WMfRdlADwyL4kZ/🔸-Snackbar-Specs?type=design&node-id=0-1&mode=design&t=TkcXVGbNbzi3HV08-0): 
1. Create a Snackbar without an action but with an indefinite length.
2. Create a Snackbar with an action and five seconds length.

@yceballost Should we allow these cases?

library/src/main/res/layout/snackbar_custom_layout_merge.xml
```diff
@@ -0,0 +1,77 @@
...
+			app:layout_constraintTop_toTopOf="parent"
+			app:layout_constraintBottom_toBottomOf="parent"
+			android:layout_marginEnd="8dp"
+			android:background="?selectableItemBackgroundBorderless"
+			android:visibility="gone"
+			tools:visibility="visible">
+		<ImageView
+				android:layout_width="match_parent"
+				android:layout_height="match_parent"
+				android:contentDescription="@string/close_button_content_description"
```
idenjoe, line 71: @yceballost @aweell According to the [specs](https://www.figma.com/file/yCHLIfy4WMfRdlADwyL4kZ/🔸-Snackbar-Specs?type=design&node-id=1038-1306&mode=design&t=jAAklkukvhlFC4lh-0) in this PR we should not include the close button which is for future development. Can you clarify it?

aweell: The dismiss action was out of the scope of this task initially. https://github.com/Telefonica/mistica-design/issues/1476

Since the snackbar action didn't perform any custom actions apart from dismiss, the user was able to dismiss the snackbar even changing the duration to infinite.

https://github.com/Telefonica/mistica-design/issues/1461 was intended as a future development. 

Since we were proposing that a custom action can be performed apart from the dismiss action. We need to ensure the user can close the snackbar, that's why the addition of the close icon.

If we're okey to merge both definitions in iOS and Web (@atabel) and follow what android already implemented I'll change the specs to include both in the same task.

I apologize for the misunderstanding @idenjoe

library/src/main/java/com/telefonica/mistica/feedback/SnackbarBuilder.kt
yamal-alm: Spec has been updated with a more clear description, I'll update this code according to it:
- No action: 5 seconds or indefinite with a dismiss x
- Action: 10 seconds or indefinite. Optional dimiss X

idenjoe (approved):

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/SnackBarCatalogFragment.kt
```diff
@@ -51,11 +55,16 @@ class SnackBarCatalogFragment : Fragment() {
...
                 inputAction.text.toString().let { actionText ->
                     if (actionText.isNotEmpty()) {
-                        withAction(actionText, { })
+                        withAction(actionText) { }
                     }
                 }
+                if (alwaysShowDismiss.isChecked()) {
+                    withDismiss()
+                }
                 val duration = when {
```
yceballost, line 64: I think you can replace "5 seconds" & "10 seconds" radio button by "Default" in catalog 🤔

- Default
- Indefinite lenght

yceballost (changes requested):

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/SnackBarCatalogFragment.kt
yamal-alm: it makes sense. Then I would replace the whole radio group with just a checkbox: ☑️  Indefinite length. WDYT

yceballost: perfect!

yceballost (approved):

PR #310, opened 2023-10-30
ANDROID-13833 Avoid problems using VivoNew in Novum

dagonco:
Be able to merge VivoNew style to Vivo in Novum.

* Avoid VivoNew to be inherit of mistica due in novum is not merging styles properly with Vivo.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

- [ ] 🖼️ Screenshots/Videos


library/src/main/res/values-night/themes_vivoNew_autogenerated.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 	<!-- Code generated automatically, DO NOT EDIT manually. Use Import design tokens GitHub Action. -->
-	<style name="MisticaTheme.VivoNew" parent="MisticaTheme.VivoNew_Customizations">
+	<!-- Important: Do NOT let this file to be style name and its parent overwritten. There is a issue merging this style with Vivo regular. -->
```
yamal-alm, line 4: this will be reverted on the next tokens auto-generation execution, right? This is something we should restrict programatticaly on the generator code

dagonco: I'm on it!

DevPabloGarcia (approved):
👍

library/src/main/res/values-night/themes_vivoNew_autogenerated.xml
dagonco: Yep, i'm trying to do that easiest as possible with no braking anything.

dagonco: Done! You could recheck that!

library/src/main/res/values/themes_vivoNew.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-	<style name="MisticaTheme.VivoNew_Customizations" parent="MisticaTheme.VivoNew_Base">
+	<style name="VivoNew_Customizations" parent="VivoNew_Base">
```
yamal-alm, line 4: hmmm, now if any app uses this VivoNew theme, it won't inherit every material attribute that it used to from `MisticaTheme`.

https://github.com/Telefonica/mistica-android/blob/32738dfdfa26add1e68b3a1151433652f827f657/library/src/main/res/values/themes.xml#L6

I know this workaround fixes our other problem when using "Vivo" with overrides from "VivoNew" but I'm starting to think that maybe the fix should be on our other app instead of in mistica, which should be agnostic about how the client uses these themes.

PR #318, opened 2023-11-16
ANDROID-14001 Fix chevron glitch on button link

yamal-alm:
There is a UI glitch when a button2 with "link" style and chevron is rendered because of how chevron height is calculated. See the following video:

https://github.com/Telefonica/mistica-android/assets/47142570/e02ac130-279c-4296-b6ae-a7a0e9a35ad6

Use built in `IntrinsicSize.Min` as the chevron height instead of listening to the height of the text which makes the chevron to switch from big to small very quickly.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [x] [Mistica App download link](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/379)
- [x] 🖼️ Screenshots

https://github.com/Telefonica/mistica-android/assets/47142570/7236b0bb-55f9-4c01-abf9-209738f50081

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
@@ -8,6 +8,7 @@ import androidx.compose.animation.slideInVertically
 import androidx.compose.animation.slideOutVertically
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/telefonica/mistica/compose/button/Button.kt
```diff
@@ -170,32 +169,31 @@ private fun ButtonContent(
...
             )
             if (withChevron) {
-                Spacer(modifier = Modifier.width(dimensionResource(id = R.dimen.button_chevron_padding)))
-                CompositionLocalProvider(
-                    LocalContentAlpha provides if (enabled) ContentAlpha.high else ContentAlpha.disabled,
+                Box(
+                    modifier = Modifier
+                        .height(IntrinsicSize.Min)
+                        .aspectRatio(CHEVRON_ASPECT_RATIO)
+                        .align(Alignment.CenterVertically)
```
yamal-alm, line 184: I needed to wrap the whole `CompositionLocalProvider` element inside a box because it doesn't allow any modifier and `IntrinsicSize` only applies to direct children of `Row`

dagonco (approved):
🪄

DevPabloGarcia (approved):
Thanks 😄

PR #321, opened 2023-11-29
Remove default padding in compose wrap button

DevPabloGarcia:
Remove default padding in compose wrap button (button2.Button)

When we use button2.Button in xml, it has a default padding. We should add the option to remove this default padding

* Add a attr to set default padding

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

No tests

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/button2/Button.kt
+++ b/library/src/main/java/com/telefonica/mistica/button2/Button.kt
@@ -85,6 +85,7 @@ class Button @JvmOverloads constructor(
                 enabled = isEnable,
                 icon = icon,
                 withChevron = withChevron,
+                invalidatePaddings = true,
                 onClickListener = onClick
             )
         }


library/src/main/java/com/telefonica/mistica/button2/Button.kt
```diff
@@ -26,6 +26,7 @@ class Button @JvmOverloads constructor(
     private var style: ButtonStyle by mutableStateOf(ButtonStyle.PRIMARY)
     private var isEnable: Boolean by mutableStateOf(true)
     private var onClick: () -> Unit by mutableStateOf({})
+    private var invalidatePadding: Boolean by mutableStateOf(false)
```
jeslat, line 29: I would rename all to `invalidatePaddings` just to maintain the same name that we use in compose

jeprubio (approved):
Nice

PR #322, opened 2023-12-18
ANDROID-14098 set asset dimensions

jeprubio:
### :tickets: Jira ticket
[ANDROID-14098](https://jira.tid.es/browse/ANDROID-14098)

Be able to set asset dimensions in ListRow.

Trying not to add breaking changes:
- Add TYPE_IMAGE_ROUNDED and the possibility to set the dimensions in the traditional views version -> [8ce1102](https://github.com/Telefonica/mistica-android/commit/8ce1102e261ba04da834fd9e0eb4360d73787310)
- Add ImageAsset to compose ListRowIcon to accept different sizes -> [68ee6ec](https://github.com/Telefonica/mistica-android/commit/68ee6ecbeea631c1b5f1122f64cf680c35ca32a3)

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

Traditional views:
TraditionalViews
Compose:
Compose
- [ ] ...


library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -283,16 +287,17 @@ class ListRowView @JvmOverloads constructor(
...
     private fun updateIconVisibility() {
         assetCircularImageView.isVisible = assetType == TYPE_IMAGE
-        assetRoundedImageView.isVisible = assetType == TYPE_IMAGE_1_1 || assetType == TYPE_IMAGE_7_10 || assetType == TYPE_IMAGE_16_9
+        assetRoundedImageView.isVisible = assetType == TYPE_IMAGE_1_1 || assetType == TYPE_IMAGE_7_10
+                || assetType == TYPE_IMAGE_16_9 || assetType == TYPE_IMAGE_ROUNDED
         assetImageView.isVisible = assetType == TYPE_SMALL_ICON || assetType == TYPE_LARGE_ICON
     }
 
-    fun setAssetType(@AssetType type: Int) {
+    fun setAssetType(@AssetType type: Int, dimensions: ImageDimensions? = null) {
```
dpastor, line 295: Appart from programatic configuration we expose also configurable elements with xml attributes. Check ListRowView styleable in `attrs_components.xml` to add new asset type enum and height/width attributes.

jeslat:
Update also the READMEs that we have for each component

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
jeprubio: Sure, I forgot doing that, I'll do that now. Thanks!

jeprubio: Added
Screenshot 2023-12-19 at 10 15 14
Screenshot 2023-12-19 at 10 17 07

library/src/main/res/values/attrs_components.xml
```diff
@@ -69,6 +69,13 @@
             <enum name="image_1_1" value="3" />
             <enum name="image_7_10" value="4" />
             <enum name="image_16_9" value="5" />
+            <enum name="image_rounded" value="6" />
+        </attr>
+        <attr name="listRowAssetHeight" format="integer">
+            <enum name="undefined" value="64" />
+        </attr>
+        <attr name="listRowAssetWidth" format="integer">
+            <enum name="undefined" value="64" />
```
yamal-alm, line 78: should these attrs be of type "dimension" in case someone uses them as follows?
```xml
listRowAssetWidth="@dimen/some_width"
```

dpastor (approved):

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
jeslat: Is it working the binding method with this new parameter? Or is it just ignored because is nullable?

library/src/main/res/values/attrs_components.xml
jeslat: +1, I think it should be `dimension` to avoid putting just a raw number

library/src/test/java/com/telefonica/mistica/compose/list/ListRowItemKtTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.telefonica.mistica.list.model.ImageDimensions
+import com.telefonica.mistica.testutils.ScreenshotsTest
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import com.telefonica.mistica.R
+
+@RunWith(RobolectricTestRunner::class)
+internal class ListRowItemKtTest: ScreenshotsTest() {
```
jeslat, line 20: Just curious, why that `Kt` in the middle of the class name?

library/src/main/res/values/attrs_components.xml
jeprubio: Yup, I can do this now. Good idea.

library/src/test/java/com/telefonica/mistica/compose/list/ListRowItemKtTest.kt
jeprubio: When you create a test with Android Studio it creates them this way sometimes, I'm not sure which rules does it follow. I can remove that if you want.

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
jeprubio: That one of the fun setAssetType will be null with binding and there are the other two of setAssetHeight and setAssetWidth to set the dimensions. That param is in order to use it when creating the element programmatically, but can be changed, it was thought as a shortcut.

library/src/main/res/values/attrs_components.xml
jeprubio: Done, both in [1c1ee2b](https://github.com/Telefonica/mistica-android/pull/322/commits/1c1ee2bab38e8e79034b55a3e8c77d22bdce06aa)

jeslat (approved):
🌭

yamal-alm (approved):
nice job 👍

PR #324, opened 2023-12-22
ANDROID-14134 Set assets clickable

jeprubio:
Until now the assets in ListRow didn't have a possible interaction different from the rest of the row but this should be allowed as it has been defined in the specs: https://www.figma.com/file/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?type=design&node-id=0%3A608&mode=design&t=lG2wUE8oNXAI5z4d-1

- Added setAssetOnClickListener for traditional views.
- Made the modifier accessible for the different ListRowIcon so they can be made clickable.
- Added tests for the traditional views and compose elements.
- Added a row example in the catalog at the end.
- Also added some section titles in the compose catalog as the traditional views already had.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 23.

In the catalog.
Classic views:
clickableAsset
Compose:
composeClickableAsset
- [ ] 🖼️ Screenshots/Videos
- [ ] ...

### DIFF
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
@@ -567,13 +567,23 @@ class ListsCatalogFragment : Fragment() {
             )
         }
 
-        override fun getItemCount(): Int = 1
+        override fun getItemCount(): Int = 2
 
         override fun onBindViewHolder(holder: ListViewHolder, position: Int) {
             with(holder.rowView) {
-                setTitle("Clickable Asset")
-                setAssetOnClickListener {
-                    Toast.makeText(context, "Asset clicked!", Toast.LENGTH_SHORT).show()
+                if (position == 0) {
+                    setTitle("Clickable Asset")
+                    setAssetOnClickListener {
+                        Toast.makeText(context, "Asset clicked!", Toast.LENGTH_SHORT).show()
+                    }
+                } else {
+                    setTitle("Clickable Asset in Clickable Row")
+                    setOnClickListener {
+                        Toast.makeText(context, "Row clicked!", Toast.LENGTH_SHORT).show()
+                    }
+                    setAssetOnClickListener {
+                        Toast.makeText(context, "Asset clicked!", Toast.LENGTH_SHORT).show()
+                    }
                 }
             }
         }
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
@@ -50,6 +50,9 @@ class ListsCatalogFragment : Fragment() {
 
         val boxedInverseList: MisticaRecyclerView = view.findViewById(R.id.boxed_inverse_list)
         boxedInverseList.adapter = ListAdapter(backgroundType = ListRowView.BackgroundType.TYPE_BOXED_INVERSE)
+
+        val clickableRow: MisticaRecyclerView = view.findViewById(R.id.clickable_list)
+        clickableRow.adapter = ClickableListAdapter(backgroundType = ListRowView.BackgroundType.TYPE_NORMAL)
     }
 
     class ListAdapter(
@@ -552,6 +555,31 @@ class ListsCatalogFragment : Fragment() {
     }
 
     class ListViewHolder(val rowView: ListRowView) : RecyclerView.ViewHolder(rowView)
+
+    class ClickableListAdapter(
+        @ListRowView.BackgroundType private val backgroundType: Int,
+    ) : RecyclerView.Adapter<ListViewHolder>() {
+        override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ListViewHolder {
+            return ListViewHolder(
+                LayoutInflater.from(parent.context).inflate(
+                    R.layout.screen_fragment_lists_catalog_item,
+                    parent,
+                    false
+                ) as ListRowView
+            )
+        }
+
+        override fun getItemCount(): Int = 1
+
+        override fun onBindViewHolder(holder: ListViewHolder, position: Int) {
+            with(holder.rowView) {
+                setTitle("Clickable Asset")
+                setAssetOnClickListener {
+                    Toast.makeText(context, "Asset clicked!", Toast.LENGTH_SHORT).show()
+                }
+            }
+        }
+    }
 
     private companion object {
         const val IMAGE_URL = "https://www.fotoaparat.cz/imgs/a/26/2639/0n1wjdf0-cr-em13-09-1200x627x9.jpg"
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
@@ -52,7 +52,7 @@ class ListsCatalogFragment : Fragment() {
         boxedInverseList.adapter = ListAdapter(backgroundType = ListRowView.BackgroundType.TYPE_BOXED_INVERSE)
 
         val clickableRow: MisticaRecyclerView = view.findViewById(R.id.clickable_list)
-        clickableRow.adapter = ClickableListAdapter(backgroundType = ListRowView.BackgroundType.TYPE_NORMAL)
+        clickableRow.adapter = ClickableListAdapter()
     }
 
     class ListAdapter(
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
```diff
@@ -110,7 +116,7 @@ sealed class ListRowIcon(val contentDescription: String?) {
             Image(
                 painter = painter,
                 contentDescription = contentDescription,
-                modifier = Modifier
+                modifier = modifier
                     .size(40.dp)
```
AnaMontes11, line 120: For accessibility reasons, we might consider expanding the clickable area of the asset to 48px. We have updated our specs.

jeprubio: OK, I'm fine with that if the specs have also been updated. Done.

AnaMontes11 (approved):
Great job!

library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
jeprubio: This is just exposing the modifier and not making it clickable with just that. Also doing that includes breaking changes in the lib. In my opinion we should ensure what we want to make clickable or not on the designs and then use this modifier to make it clickable or not depending on those designs. I'm going to revert this change.

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -309,6 +309,10 @@ class ListRowView @JvmOverloads constructor(
         }
     }
 
+    fun setAssetOnClickListener(clickListener: OnClickListener) {
+        assetImageLayout.setOnClickListener(clickListener)
+    }
```
jeprubio, line 314: This is the code added for classic views in order to be able to set the on click listener.

library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
```diff
@@ -28,31 +28,36 @@ sealed class ListRowIcon(val contentDescription: String?) {
     data class NormalIcon(
         val painter: Painter? = null,
         private val description: String? = null,
+        val modifier: Modifier = Modifier,
```
jeprubio, line 31: In compose views we expose the modifier which is a common used practice which we were not following until now: https://chrisbanes.me/posts/always-provide-a-modifier/

library/src/test/java/com/telefonica/mistica/list/ListRowViewTest.kt
```diff
@@ -26,4 +27,20 @@ internal class ListRowViewTest: ScreenshotsTest() {
...
+            activity.layoutInflater.inflate(R.layout.test_list_row_view, wrapper, true)
+
+            with (activity.findViewById<ListRowView>(R.id.list_row_view_32)) {
+                setOnClickListener { clicks++ }
+                performClick()
+            }
+
+            assertEquals(1, clicks)
+        }
+    }
```
jeprubio, line 45: Tests that the asset is clickable on xml

library/src/test/java/com/telefonica/mistica/compose/list/ListRowItemKtTest.kt
```diff
@@ -35,14 +43,30 @@ internal class ListRowItemKtTest: ScreenshotsTest() {
...
+            clicked++
+        }
+        `when ListRowItem with asset`(
+            dimensions = ImageDimensions(width = 44, height = 44),
+            onAssetClick = onAssetClick,
+        )
+        composeTestRule.onNode(hasTestTag(LIST_ROW_ITEM_ASSET_TAG)).performClick()
+
+        assertEquals( 1, clicked)
+    }
```
jeprubio, line 59: Tests that the asset is clickable on compose

library/src/test/java/com/telefonica/mistica/compose/list/ListRowItemKtTest.kt
```diff
@@ -35,14 +43,30 @@ internal class ListRowItemKtTest: ScreenshotsTest() {
...
+            clicked++
+        }
+        `when ListRowItem with asset`(
+            dimensions = ImageDimensions(width = 44, height = 44),
+            onAssetClick = onAssetClick,
+        )
+        composeTestRule.onNode(hasTestTag(LIST_ROW_ITEM_ASSET_TAG)).performClick()
+
+        assertEquals(1, clicked)
+    }
```
jeprubio, line 59: Tests that the asset is clickable on compose

library/src/test/java/com/telefonica/mistica/list/ListRowViewTest.kt
```diff
@@ -26,4 +27,20 @@ internal class ListRowViewTest: ScreenshotsTest() {
             compareScreenshot(Espresso.onView(ViewMatchers.withId(R.id.dummy_activity_wrapper)))
         }
     }
+
+    @Test
+    fun `check ListRowView xml onClick`() {
+        rule.scenario.onActivity { activity ->
+            var clicks = 0
```
jdelga, line 34: Same here, it is used as a boolean

jdelga (approved):
Cool! Good job!

library/src/test/java/com/telefonica/mistica/compose/list/ListRowItemKtTest.kt
```diff
@@ -35,14 +43,30 @@ internal class ListRowItemKtTest: ScreenshotsTest() {
         `then screenshot is OK`()
     }
 
+    @Test
+    fun `check ListRowItem with clickable asset`() {
+        var clicked = 0
```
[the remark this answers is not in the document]
jeprubio: Not exactly as I want to know that it has been invoked only once. A boolean would only ensure it has been invoked but it could be multiple times

library/src/test/java/com/telefonica/mistica/list/ListRowViewTest.kt
jeprubio: Same answer here

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Lists.kt
```diff
@@ -298,10 +302,14 @@ const val IMAGE_URL = "https://www.fotoaparat.cz/imgs/a/26/2639/0n1wjdf0-cr-em13
 @Composable
 fun Lists() {
     val samples = samples()
+    val context = LocalContext.current
```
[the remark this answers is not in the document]
jeprubio: but can't ve moved to the same line it's used as that would not compile:
Screenshot 2023-12-26 at 12 07 00

library/src/test/java/com/telefonica/mistica/compose/list/ListRowItemKtTest.kt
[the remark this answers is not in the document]
jdelga: The condition is exactly the same, you are checking if it has been clicked 1 time instead of 0.

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
```diff
@@ -553,6 +556,31 @@ class ListsCatalogFragment : Fragment() {
 
     class ListViewHolder(val rowView: ListRowView) : RecyclerView.ViewHolder(rowView)
 
+    class ClickableListAdapter(
+        @ListRowView.BackgroundType private val backgroundType: Int,
```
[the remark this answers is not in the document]
jeprubio: True, it's not needed, removed from here the catalog. Thanks.

catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
```diff
@@ -553,6 +556,29 @@ class ListsCatalogFragment : Fragment() {
...
+                ) as ListRowView
+            )
+        }
+
+        override fun getItemCount(): Int = 1
+
+        override fun onBindViewHolder(holder: ListViewHolder, position: Int) {
+            with(holder.rowView) {
+                setTitle("Clickable Asset")
+                setAssetOnClickListener {
```
dagonco, line 575: From the way I have understood Figma, the list row should be clickable.
Otherwise, probably is a good idea to have all or atleast some of the list row variants with this new modification.
This is the intention of the catalog. A sample of the capabilities that each Mística's component has.

jeprubio: Added an extra row which is clickable the row and the asset to each of the catalogs if it's that what you mean.

dagonco (approved):
Niiice!

PR #326, opened 2024-01-09
ANDROID-14153 Update AGP, target SDK, Kotlin and Compose

yamal-alm:
- Update AGP to 8
- Update target sdk to 34
- Update kotlin to 1.9
- Update Compose to latest version

- Increase AGP version to 8.1.4
  - Move `DummyActivity` from `library/test` to a new `library-test-utils` module because now Activities under test folder are not included in the merged manifest and they need to be defined inside a valid build type. This module is only included as a `testImplementation` inside `library` module and by now it only contains this `DummyActivity`
- Increase target sdk to 34 and do required migration steps
  - Bump java version to from 8 to 17 
  - Setting up java version to 17 on Github workflows
  - Explicitly enable buildConfig fields
  - Remove unnecessary gradle config flags: `r8.failOnMissingClasses` and `databinding.incremental`
  - Add needed gradle config flag `android.nonTransitiveRClass=false`
  - Rename some deprecated gradle methods  their new signatures
- Update Kotlin to 1.9.21
- Update compose version to latest version
- Update accompanist to latest stable version (0.32.0) 

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,7 +10,7 @@ _Provide a description of the implementation. A list of steps would be ideal._
 ### ☑️ Checks
 - [ ] I updated the documentation, including readmes and wikis. If this is a breaking change, update [UPGRADING.md](../UPGRADING.md) to inform users how to proceed. If no updates are necessary, indicate so.
 - [ ] Tested with dark mode.
-- [ ] Tested with API 23.
+- [ ] Tested with API 21.
```
yamal-alm, line 13: min sdk is still 21

.github/workflows/catalog.yml
```diff
@@ -8,16 +8,22 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-    - name: Check pull request status
-      if: "!github.event.pull_request.merged == true"
-      run: 'echo "Skipping generation, pull request is not merged"'
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: '17'
```
yamal-alm, line 15: this is the only new step, other changes are just formatting

.github/workflows/debug_catalog.yml
```diff
@@ -15,13 +15,19 @@ jobs:
...
+          java-version: '17'
+
       - name: Checkout repo
         uses: actions/checkout@v3
 
       - name: Publish catalog
         env:
           APPCENTER_API_TOKEN: ${{ secrets.APPCENTER_API_TOKEN }}
-        run: 'bash ./gradlew clean check appCenterAssembleAndUploadDebug -Dappcenter_app_name=Mistica-Catalog -Dappcenter_notify=false 
+        run: 'bash ./gradlew clean appCenterAssembleAndUploadDebug -Dappcenter_app_name=Mistica-Catalog -Dappcenter_notify=false
```
yamal-alm, line 30: there is no need to run `check` task on the debug catalog generation workflow

.github/workflows/tests.yml
```diff
@@ -8,19 +8,23 @@ concurrency:
 
 jobs:
   tests:
-    runs-on: self-hosted-novum
+    runs-on: ubuntu-latest
```
yamal-alm, line 11: by now, we are not launching any emulator, so we can rely on ubuntu-latest runners from Github Actions

.github/workflows/tests.yml
```diff
@@ -8,19 +8,23 @@ concurrency:
...
-    
-    - name: "Build Android project"
-      uses: docker://docker.tuenti.io/android/novum_android_api31:2
-      with:
-        args: 'bash ./gradlew clean check assemble'
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: '17'
```
yamal-alm, line 17: this is the only new step, the other changes are just formatting

build.gradle
```diff
@@ -1,13 +1,14 @@
...
-        kotlin_version = "1.8.21"
+        min_sdk_version = 21
+        target_sdk_version = 34
+        compose_ui_version = "1.5.4"
+        compose_compiler_version = '1.5.7'
+        kotlin_version = "1.9.21"
         androidx_appcompat_version = "1.6.0"
         androidx_activity_compose_version = "1.6.0"
-        accompanist_version = "0.30.1"
+        accompanist_version = "0.32.0"
```
yamal-alm, line 11: main changes in this PR:
- kotlin to 1.9.21
- compose compiler to 1.5.7 (according to [this mapping](https://developer.android.com/jetpack/androidx/releases/compose-kotlin)) and compose to 1.5.4 (latest)
- AGP to 8.1.4
- Accompanist to 0.32.0 (with support for compose 1.5.X)

library-test-utils/build.gradle
```diff
@@ -0,0 +1,43 @@
+plugins {
```
yamal-alm, line 1: `DummyActivity` has been moved from `library/test` to a new `library-test-utils` module because now Activities under test folder are not included in the merged manifest and they need to be defined inside a valid build type. This module is only included as a `testImplementation` inside `library` module and by now it only contains this `DummyActivity`.

Related link https://github.com/robolectric/robolectric/pull/4736

library/src/main/java/com/telefonica/mistica/input/TextInput.kt
```diff
@@ -261,32 +260,5 @@ class TextInput @JvmOverloads constructor(
...
-                    s: CharSequence?,
-                    start: Int,
-                    count: Int,
-                    after: Int
-                ) {
-                }
-
-                override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {}
-            }
-            val oldTextWatcher = ListenerUtil.trackListener(input, textWatcher, R.id.textWatcher)
```
yamal-alm, line 285: this `textWatcher` id is no longer available, maybe because it is something internal that we shouldn't have been relying on. After a quick search, it looks like this method is not being used at all by anyone

gradle.properties
```diff
@@ -17,8 +17,4 @@ android.useAndroidX=true
 
 kapt.incremental.apt=true
 
-# This will be default on AGP 8.0
-android.r8.failOnMissingClasses=true
-
-###### Project settings ######
-android.databinding.incremental=true
+android.nonTransitiveRClass = false
```
yamal-alm, line 20: this keeps allowing us to import R class just from current module and still referencing resources from others. With new AGP and without this flag, we would have to import different R classes.

gradle.properties
```diff
@@ -17,8 +17,4 @@ android.useAndroidX=true
 
 kapt.incremental.apt=true
 
-# This will be default on AGP 8.0
-android.r8.failOnMissingClasses=true
-
-###### Project settings ######
-android.databinding.incremental=true
```
yamal-alm, line 24: these flags are deprecated in AGP 8 (also they are true by default)

library-test-utils/build.gradle
jeprubio: That's going to conflict with the other currently opened pr https://github.com/Telefonica/mistica-android/pull/327

yamal-alm: thanks, good to know! But can you point me out which file is going to generate conflicts? The package of DummyActivity hasn't changed so every usage of it shouldn't be a problem 🤔

jeprubio: It's not a merge conflict, I was thinking of the library/src/test/res/layout/test_mistica_recycler_view.xml which my understanding is that this should also not be in /test/res or is there no problem with the resources in test layouts and it's only about activities? If it's only about activities then there should be no problem

jeprubio (approved):
LGTM thanks 👍

library-test-utils/build.gradle
yamal-alm: only the Activity got affected, resources can be located inside library module!

.github/workflows/catalog.yml
jeslat: no JDK 17 in ubuntu-latest by default?

jeslat: In theory is already installed https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#java

yamal-alm: before adding this line, I was getting an error about java 17 version not being used. Let me try to execute `java --version` as the first step in the workflow and see what is its output

yamal-alm: I have tried to another execution without these lines and I got this error where it says that we need to use Java 17 because Java 11 is being used by default
https://github.com/Telefonica/mistica-android/actions/runs/7477643664/job/20350791222?pr=326

jeslat: Ok, thanks for checking it, I guess it installed but it's not the default version

jeslat (approved):
good job!

app/build.gradle
```diff
@@ -8,12 +8,13 @@ def appCenterProjects = new Properties()
 file("appcenter_projects.properties").withInputStream { appCenterProjects.load(it) }
 
 android {
-    compileSdkVersion 33
+    compileSdk target_sdk_version
```
jdelga, line 11: Compile sdk is not the same as target sdk

yamal-alm: yes, I know. But given that both are always updated at the same time, I opted to reuse the same variable

jdelga: Actually I don't think they should be done at the same time, but that's why I have only left it as a comment. Just in case you missed it.

PR #327, opened 2024-01-10
Android 13880 Hide last divider in XML Lists

pmartinbTEF:
Hide last divider by default in xml lists.

In Compose elements and [dividers are added manually](https://github.com/Telefonica/mistica-android/blob/main/library/src/main/java/com/telefonica/mistica/compose/list/README.md#divider) so no changes needed

Check is last element in DividerItemDecoration

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 23.

- [x] Create a MisticaRecyclerView in xml
- [x] Check that last divider is hidden
- [x] Added Screenshot test


library/src/test/res/layout/test_mistica_recycler_view.xml
```diff
@@ -0,0 +1,17 @@
...
+		android:padding="16dp"
+		android:orientation="vertical">
+	<com.telefonica.mistica.list.MisticaRecyclerView
+			android:id="@+id/test_list"
+			android:layout_width="match_parent"
+			android:layout_height="wrap_content"
+			android:layout_marginTop="8dp"
+			android:nestedScrollingEnabled="false"
+			app:listLayoutType="full_width"
+			/>
```
jeprubio, line 16: Perhaps this could also be created by code and inserted in the dummy layout. But I'm not against this, if we need to test several recycler views perhaps it's simpler this way.

jeprubio (approved):
Nice approach 👏  check if the new screenshot should also be in this pr.

library/src/test/java/com/telefonica/mistica/list/MisticaRecyclerViewTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    fun `check MisticaRecyclerView xml hides last divider`() {
+        rule.scenario.onActivity { activity ->
+            val wrapper: FrameLayout = activity.findViewById(R.id.dummy_activity_wrapper)
+            val view = activity.layoutInflater.inflate(R.layout.test_mistica_recycler_view, wrapper, false)
+            activity.setContentView(view)
+
+            val list: MisticaRecyclerView = view.findViewById(R.id.test_list)
+            list.adapter = ClickableListAdapter()
+            compareScreenshot(Espresso.onView(ViewMatchers.withId(R.id.test_recycler_view)))
```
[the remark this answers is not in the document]
pmartinbTEF: Done!

jeslat (approved):
👍

library/src/test/res/layout/test_mistica_recycler_view.xml
pmartinbTEF: I think this approach is simpler and, as you said, the layout is only added to the test directory :)

PR #330, opened 2024-02-13
ANDROID-14295 Support urls in images in Sheet

jeslat:
Add support to load images from a URL in the Sheet component

* To not break the compatibility, I'm adding a new parameter `RowAsset` which supports the different image types and deprecating the old one.
* At the visual level, nothing has changed. 
* To load the URL I'm using `coil` library. To support SVGs, I've added `coil-svg`.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

- It can be tested in the catalog app enabling the new checkbox `Image from url`

drawing


library/src/main/java/com/telefonica/mistica/sheet/SheetModel.kt
```diff
@@ -66,16 +68,50 @@ data class ActionButton(
...
+    ) : InformativeIcon()
+
+    data class SmallIcon(
+        @Deprecated("Use new field rowAsset. RowAsset will have preference over asset")
+        val drawableRes: Drawable? = null,
+        val rowAsset: RowAsset? = null,
+    ) : InformativeIcon()
+}
+
+sealed class RowAsset {
```
jeslat, line 111: New model which supports the three types of image source

library/src/main/java/com/telefonica/mistica/util/Base64DataUrlFetcher.kt
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Coil Fetcher implementation for Data Urls encoded in Base64.
+ * Format: data:mime/type;param1=value1;paramN=valueN;base64,actual_data
+ * In the original spec, everything is optional except "data:" and ",actual_data".
+ * But in this particular implementation, ";base64" appearance is also mandatory.
+ * The appearance of ";base64" means that the data is encoded as base64,
+ * otherwise the data would be represented using ASCII encoding.
+ * Link to Spec: https://datatracker.ietf.org/doc/html/rfc2397
+ */
+class Base64DataUrlFetcher(
```
jeslat, line 25: Copy/paste from android-messeger to support SVGs encoded in base64

library/src/main/java/com/telefonica/mistica/sheet/SheetModel.kt
```diff
@@ -66,16 +68,50 @@ data class ActionButton(
...
 sealed class SelectableAsset {
-    data class Image(val drawableRes: Drawable) : SelectableAsset()
-    data class SmallImage(val drawableRes: Drawable) : SelectableAsset()
-    data class SmallIcon(@DrawableRes val id: Int) : SelectableAsset()
-    data class LargeIcon(@DrawableRes val id: Int) : SelectableAsset()
+    data class Image(
+        @Deprecated("Use new field rowAsset. RowAsset will have preference over asset")
+        val drawableRes: Drawable? = null,
+        val rowAsset: RowAsset? = null,
+    ) : SelectableAsset()
```
yamal-alm, line 75: Can you check if the following suggestion works? I think this can save us from having to deal with nullable fields and their corresponding `IllegalArgumentException`s:
```kotlin
data class Image(
    val rowAsset: RowAsset? = null,
) : SelectableAsset() {
    @Deprecated("Use primary constructor providing a RowAsset object instead")
    constructor(drawableRes: Drawable) : this(RowAsset.DrawableAsset(drawableRes))
}

```

This should also apply for `InformativeIcon` types

library/src/main/java/com/telefonica/mistica/util/ImageViewUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+        is RowAssetViewData.DrawableIdAsset -> this.setImageDrawable(AppCompatResources.getDrawable(this.context, rowAsset.id))
+        is RowAssetViewData.UrlAsset -> this.loadUrl(rowAsset.url)
+    }
+}
+
+fun ImageView.loadUrl(
+    url: String,
+    builder: ImageRequest.Builder.() -> Unit = {},
+) {
+    val imageLoader = ImageLoader.Builder(context)
```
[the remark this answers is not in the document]
jeslat: I've created a singleton using an object to provide just one `ImageLoader` https://github.com/Telefonica/mistica-android/pull/330/commits/17aa31c363198aa4855ff2f020387437b9b55fa8

library/src/main/java/com/telefonica/mistica/sheet/SheetModel.kt
```diff
@@ -66,16 +68,56 @@ data class ActionButton(
...
+    ) : SelectableAsset() {
+        @Deprecated("Use primary constructor providing a RowAsset object instead")
+        constructor(drawableRes: Drawable) : this(RowAsset.DrawableAsset(drawableRes))
+    }
+
+    data class SmallIcon(
+        val rowAsset: RowAsset,
+    ) : SelectableAsset() {
+        @Deprecated("Use primary constructor providing a RowAsset object instead")
+        constructor(id: Int) : this(RowAsset.DrawableIdAsset(id))
```
yamal-alm, line 89: should we add @DrawableRes annotation as well?

library/src/main/java/com/telefonica/mistica/sheet/children/list/ListElementViewData.kt
```diff
@@ -31,27 +31,33 @@ internal sealed class ListElementViewData(
...
+    data class SmallIcon(val asset: RowAssetViewData) : InformativeIconViewData()
 }
 
 interface OnClickListener {
     fun onClicked(id: String)
-}
\ No newline at end of file
+}
+
+sealed class RowAssetViewData {
```
yamal-alm, line 59: internal?

library/src/main/java/com/telefonica/mistica/util/ImageViewUtils.kt
```diff
@@ -0,0 +1,28 @@
...
+        is RowAssetViewData.UrlAsset -> this.loadUrl(rowAsset.url)
+    }
+}
+
+fun ImageView.loadUrl(
+    url: String,
+    builder: ImageRequest.Builder.() -> Unit = {},
+) {
+
+    val request = ImageRequest.Builder(context)
```
yamal-alm, line 21: Won't this implementation leak the context of the first loaded image? I think by now we can only have one `ImageBuilder` per `Sheet` and use its context to build it. However we will maybe have to pass it down all these mappers as parameter. 

One solution can be to wrap every mapper into a single `SheetMapper` class (instantiated manually inside `Sheet`) and declare the `ImageBuilder` there as a top level reusable property. Probably a bit overengeneering xD so feel free to do what you like the most!

yamal-alm (commented):
nice! just a copuple of comments!

library/src/main/java/com/telefonica/mistica/sheet/SheetModel.kt
jeslat: ouch yes

library/src/main/java/com/telefonica/mistica/sheet/children/list/ListElementViewData.kt
jeslat: Changed, also `SelectableListAssetViewData` and `InformativeIconViewData`

haynlo (approved):
Nice job! 👍🏼

library/src/main/java/com/telefonica/mistica/util/ImageViewUtils.kt
jeslat: Talked privately, I've changed it to use the application context

yamal-alm (approved):
nice job 👏

PR #334, opened 2024-03-06
ANDROID-14416 Allow tint for RowItem icons

haynlo:
Currently the ListRowItem component in Mistica only admits a "ListRowIcon" param where we can only define a painterResource. This limitation causes that Icon types are tinted with default color (Black) instead of its original color.

We need to add a new optional param to allow original tint (tint= Color.Unspecified) or a custom tint color.

* Create new optional parameter "tint: Color? = null" for Icon builders data classes.
* Use this value when composing the Icon.
  * If it's not provided (null) use the default color keeping the same behavior as before.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

Captura de pantalla 2024-03-06 a las 16 00 39

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
@@ -9,6 +9,8 @@ import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.Icon
+import androidx.compose.material.LocalContentAlpha
+import androidx.compose.material.LocalContentColor
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -29,6 +31,7 @@ sealed class ListRowIcon(val contentDescription: String?) {
         val painter: Painter? = null,
         private val description: String? = null,
         val modifier: Modifier = Modifier,
+        val tint: Color? = null,
     ) : ListRowIcon(description)
 
     data class CircleIcon(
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
```diff
@@ -88,7 +92,8 @@ sealed class ListRowIcon(val contentDescription: String?) {
                 Icon(
                     painter = painter,
                     modifier = Modifier.size(24.dp),
-                    contentDescription = contentDescription
+                    contentDescription = contentDescription,
+                    tint = tint ?: LocalContentColor.current.copy(alpha = LocalContentAlpha.current)
```
haynlo, line 96: https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#Icon(androidx.compose.ui.graphics.ImageBitmap,kotlin.String,androidx.compose.ui.Modifier,androidx.compose.ui.graphics.Color)

library/src/main/java/com/telefonica/mistica/compose/list/ListRowIcon.kt
```diff
@@ -104,7 +109,8 @@ sealed class ListRowIcon(val contentDescription: String?) {
            painter?.let {
                 Icon(
                     painter = painter,
-                    contentDescription = contentDescription
+                    contentDescription = contentDescription,
+                    tint = tint ?: LocalContentColor.current.copy(alpha = LocalContentAlpha.current)
```
jeprubio, line 113: Perhaps you could add trailing commas to tint so that it won't happen as in here with the contentDescription in a new pr if another parameter is added. And the same in the other change.

jeprubio (approved):
👍

DevPabloGarcia (approved):
Nice !

PR #341, opened 2024-04-04
ANDROID-14391 Do not use custom test styles

yamal-alm:
We have several ad-hoc resources used by Roborazzi tests: e.g.: MisticaTheme.Blau.test. 

These styles and other resources are exposed outside the lib. That is, they are visible to whoever uses Mistica. This is not a security problem at all but it's not a desired behavior.

- Remove following lines from library's build.gradle:
```groovy
-tasks.withType(Test) {
-    android.sourceSets.main.res.srcDirs += 'src/test/res'
-}
```
- Remove custom test styles
- Create a wrapper layout for every button type using corresponding and actual Mistica styles and inflate them accordingly from ButtonTest class
- Fix Brand mapping where Vivo style was being used for O2 and Blau
- Regenerate screenshots baseline for O2 and Blau with correct style

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

- [x] No custom styles anymore and no custom test resources exposed outside library


haynlo (approved):
Looks good, nice catch with the Brand fix!

pmartinbTEF (approved):
Great job!

PR #345, opened 2024-04-16
ANDROID-14548 update close translation

jeprubio:
Update string used for closing the bottom sheet as content description.
- Asking for the translation and updating it.
The translations:
es_es: Cerrar ventana
de_de: Fenster schließen
en_us: Close window
pt_br: Fechar janela

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 21.

- [x] 🖼️ Screenshots/Videos

(turn the volume ON)

https://github.com/Telefonica/mistica-android/assets/13270085/6adbb667-10e5-4f6f-9264-5e888b3659d4

### DIFF
--- a/library/src/main/res/values-pt/strings_content_descriptions.xml
+++ b/library/src/main/res/values-pt/strings_content_descriptions.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="close_button_content_description" comment="Content description for image views that act as buttons for dismissing views">
-Fechar janela</string>
+    <string name="close_button_content_description" comment="Content description for image views that act as buttons for dismissing views">Fechar janela</string>
     <string name="badge_notification_description" comment="Content description for non numeric badge">Novos conteúdos</string>
 </resources>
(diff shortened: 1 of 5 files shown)


library/src/main/res/values-pt/strings_content_descriptions.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="close_button_content_description" comment="Content description for image views that act as buttons for dismissing views">
-Fechar</string>
+Fechar janela</string>
```
yamal-alm, line 4: I'm not sure if this token with an initial line break is being treated as one single line or two 🤔 just thinking aloud, maybe we should align this

yamal-alm (approved):
nice 👍

library/src/main/res/values-pt/strings_content_descriptions.xml
jeprubio: Ups, it's true there is a line break, I'll remove it just in case.

pmartinbTEF (approved):
LGTM

PR #348, opened 2024-04-24
ANDROID-14573 is list row item title heading

jeprubio:
Add the possibility that a listRowItem title is set as a heading for accessibility.

- Add isTitleHeading to ListRowItem in compose [d490237](https://github.com/Telefonica/mistica-android/commit/d490237e2bf025e337b8c8a00e8c2f7eb617a2c7) - Problem: it only reads the heading for non clickable rows.
- Add isTitleHeading to ListRowView for classic views [4dcc5e6](https://github.com/Telefonica/mistica-android/commit/4dcc5e6a236d07667b3a5bac0e3c977e90e3fe17) - Problem: it only works if setting importantForAccessibility to none for that row.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

https://github.com/Telefonica/mistica-android/assets/13270085/24d548cd-6a73-4c63-bb85-aef5046ac60e


library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -38,12 +40,13 @@ import com.telefonica.mistica.compose.tag.Tag
 import com.telefonica.mistica.compose.theme.MisticaTheme
 import com.telefonica.mistica.compose.theme.brand.MovistarBrand
 
-@ExperimentalMaterialApi
+@OptIn(ExperimentalMaterialApi::class)
```
jeprubio, line 43: Not using OptIn made anywhere where ListRowItem was used to have ExperimentalMaterialApi annotation

yamal-alm (approved):
nice 👍

haynlo (approved):
Nice job!! 🚀

PR #350, opened 2024-04-26
ANDROID-14597 title view heading

jeprubio:
The title of TitleView component should be able to be set with accessibility heading.

- Add isTitleHeading to TitleView
- Add isTitleHeading parameter to Title composable
- Add few samples in the catalog.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 21.

https://github.com/Telefonica/mistica-android/assets/13270085/0cc6d8c6-9fb9-4e40-8e37-0e4bf754678a
- [x] Mistica App QR or [download link](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/481)

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/title/TitleView.kt
+++ b/library/src/main/java/com/telefonica/mistica/title/TitleView.kt
@@ -30,6 +30,11 @@ import com.telefonica.mistica.util.setTextPreset
     ),
     BindingMethod(
         type = TitleView::class,
+        attribute = "isTitleHeading",
+        method = "setIsTitleHeading"
+    ),
+    BindingMethod(
+        type = TitleView::class,
         attribute = "link",
         method = "setLink"
     ),
--- a/library/src/main/java/com/telefonica/mistica/title/TitleView.kt
+++ b/library/src/main/java/com/telefonica/mistica/title/TitleView.kt
@@ -7,7 +7,9 @@ import android.widget.TextView
 import androidx.annotation.AttrRes
 import androidx.annotation.IntDef
 import androidx.annotation.StyleRes
+import androidx.annotation.VisibleForTesting
 import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.core.view.ViewCompat
 import androidx.databinding.BindingMethod
 import androidx.databinding.BindingMethods
 import com.telefonica.mistica.R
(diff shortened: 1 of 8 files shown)


jdelga (approved):

library/src/main/java/com/telefonica/mistica/title/TitleView.kt
```diff
@@ -95,6 +107,10 @@ class TitleView @JvmOverloads constructor(
         titleTextView.isAllCaps = styleConfig.isAllCaps
     }
 
+    fun setTitleHeading() {
```
yamal-alm, line 110: maybe we should add support for databinding as well and review if it was added too in latest changes to ListRowItem

library/src/main/java/com/telefonica/mistica/title/TitleView.kt
```diff
@@ -79,6 +93,9 @@ class TitleView @JvmOverloads constructor(
 
     fun getTitle(): String = titleTextView.text.toString()
 
+    @VisibleForTesting
```
yamal-alm, line 96: minor: can we use findViewById instead of exposing this method?

yamal-alm (approved):
nice 👍 just a small suggestion

library/src/main/java/com/telefonica/mistica/title/TitleView.kt
jeprubio: It has only internal visibility, we could but as it's not added by a layout and instead is added programmatically:
````
val wrapper: FrameLayout = activity.findViewById(R.id.dummy_activity_wrapper)
val titleView = TitleView(activity)
wrapper.addView(titleView)
```
 I don't know how to do that in this case, it would if we create a layout with an id and then not create it this way. But I didn't feel necessary to create a layout for testing just for that, idk.

yamal-alm: titleTextView has an actual id (R.id.title_text) that is part of `R.layout.title`. Then in the test you would only need to invoke dummyActivity.findViewById<TextView>(R.id.title_view)

Anyways, don't change anything, I missed the `internal` modifier. It is ok 👍

jeprubio: Oh, yup, that might work as well, it wasn't what I was thinking about. But yup, it's only internal...

PR #354, opened 2024-05-16
ANDROID-14618 Checkbox links not underlined

pmartinbTEF:
Links in checkbox text should not show an underline as it can be seen in Figma

For the compose implementation I added a "underline" parameter to TextWithLinks class. This param is true by default and false for checkbox text.
In xml version links are added manually so I just updated the catalog example

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

- [x] Execute catalog app
- [x] go to Input
- [x] Check that checkbox links are not underlined

Before:
drawing

drawing

After:
drawing

drawing


library/src/main/java/com/telefonica/mistica/compose/input/TextWithLinks.kt
```diff
@@ -20,9 +20,10 @@ internal fun TextWithLinks(
     text: String,
     links: List<TextLink>,
     modifier: Modifier = Modifier,
+    underline: Boolean = true,
```
yamal-alm, line 23: since this is an internal class, do we really need to set a default value for `underline`?  Is there any scenario where we want `underline` to be set to `true`? If no, I think we can just simply set `TextDecoration.None` in `textDecoration `param bellow

library/src/main/java/com/telefonica/mistica/compose/input/TextWithLinks.kt
```diff
@@ -83,23 +87,29 @@ private fun AnnotatedString.Builder.highlight(
...
 ) {
     withStyle(
         style = SpanStyle(
             color = highlightColor,
-            textDecoration = TextDecoration.Underline,
+            textDecoration = if(underline) {
+                TextDecoration.Underline
+            } else {
+                TextDecoration.None
+            },
```
jeprubio, line 112: Nice 🚀 

JepLint: Just missing a space between the if and the rounded bracket 😄

library/src/main/java/com/telefonica/mistica/compose/input/TextWithLinks.kt
jeprubio, line 23: Should the default value be true if in figma the default behaviour is not having it underlined?

jeprubio (approved):
Great job 👏

library/src/main/java/com/telefonica/mistica/compose/input/TextWithLinks.kt
jeprubio: Sorry, didn't read this before, I thought it was a good idea to leave it as a parameter just defaulting to false but as you want.

yceballost (approved):
noice!

library/src/main/java/com/telefonica/mistica/compose/input/TextWithLinks.kt
pmartinbTEF: Deleted underline param and set TextDecoration.None because it is no used anywhere else

pmartinbTEF: see previous comment

library/src/main/java/com/telefonica/mistica/compose/input/TextWithLinks.kt
pmartinbTEF: Deleted :)

yamal-alm (approved):
nice job 👏

PR #355, opened 2024-05-21
ANDROID-13801 Improve buttons accessibility

jeslat:
Improve button component accessibility. 

By default, the content description of a button is the text, but we need to give support to add a custom one in our custom implementations.

* Progress button (XML):
  * It's a custom view with several buttons and views internally, we have to configure which is the important button for accessibility and ignore the others.
* Button2 (XML using compose internally):
  * We have to ignore the custom view and pass the content description to the compose component that we use internally.
* Button (compose):
  * Add a new parameter to customize the content description.
  * Use this custom content description in the internal Text.

### ☑️ Checks

- You can review the result in this video (audio on):

https://github.com/Telefonica/mistica-android/assets/2582348/00f79bdf-92a8-4e3f-b8e7-6e9d8348e934

### DIFF
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Buttons.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Buttons.kt
@@ -71,6 +71,7 @@ private fun Buttons(
                     name = it.name.lowercase().replaceFirstChar(Char::titlecase),
                     style = it,
                     isLoading = false,
+                    contentDescription = "Custom content description",
                 )
 
                 CatalogButton(
@@ -86,6 +87,7 @@ private fun Buttons(
                     style = it,
                     isLoading = isLoading,
                     loadingText = "Loading",
+                    contentDescription = "Custom content description loading",
                     onClickListener = {
                         isLoading = true
                         scope.launch {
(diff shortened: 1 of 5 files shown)


catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Buttons.kt
```diff
@@ -71,6 +71,7 @@ private fun Buttons(
                     name = it.name.lowercase().replaceFirstChar(Char::titlecase),
                     style = it,
                     isLoading = false,
+                    contentDescription = "Custom content description",
```
jeslat, line 74: I've added some custom content descriptions in the catalog to test it

library/src/main/java/com/telefonica/mistica/button/ProgressButton.kt
```diff
@@ -108,13 +111,15 @@ class ProgressButton : FrameLayout {
...
                 }
+            importantForAccessibility = IMPORTANT_FOR_ACCESSIBILITY_NO
             isIndeterminate = true
             indeterminateTintMode = PorterDuff.Mode.SRC_IN
             visibility = View.INVISIBLE
         }
 
         buttonBackground.apply {
             id = NO_ID
+            importantForAccessibility = IMPORTANT_FOR_ACCESSIBILITY_YES
```
jeslat, line 122: The only important thing for accessibility is this button which has the real text and the real state of the progress button.

yamal-alm (approved):
nice job 👏

DevPabloGarcia (approved):
Thanks !

aweell (approved):

PR #359, opened 2024-05-28
ANDROID-14716 - Add space to progress bar in button

DevPabloGarcia:
Don't cut text in loading state in buttons

If the text of the normal state and the loading state in a link button is the same, when we are showing the loading button, the text doesn't fit, showing three dots.

❗ There is only one file changed, Button.kt. The rest are baseline updates ❗ 

* When calculate button width, add progress bar size and space

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

- [x] 🖼️ Screenshots/Videos

https://github.com/Telefonica/mistica-android/assets/42326032/a072ed3e-018f-4df7-9cde-3f430bfb6a3c


pmartinbTEF (approved):
LGTM

jdelga (commented):
I think that the only updated screenshot should be the one that is a loading button (has a Loading text)

As you can see in the generated screenshots, all buttons have been updated to have a higher width. However, none of the screenshots belong to "Loading Button", as it is not defined in the tested button parameters:

https://github.com/Telefonica/mistica-android/blob/432fb54363ad0252eabe4b79f482d56f67b8649b/library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt#L48

I think that the bounding box that should be updated is only for the case of a loading button, which is the one that has a secondary text, which causes the bug.

https://github.com/Telefonica/mistica-android/blob/432fb54363ad0252eabe4b79f482d56f67b8649b/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt#L63

WDYT ?

Adit: 

Also, I think the case which is the one that has a loading text larger than standard text is not covered yet:

loading_true

DevPabloGarcia:
> I think that the only updated screenshot should be the one that is a loading button (has a Loading text)
> 
> As you can see in the generated screenshots, all buttons have been updated to have a higher width. However, none of the screenshots belong to "Loading Button", as it is not defined in the tested button parameters:
> 
> https://github.com/Telefonica/mistica-android/blob/432fb54363ad0252eabe4b79f482d56f67b8649b/library/src/test/java/com/telefonica/mistica/compose/button/ButtonKtTest.kt#L48
> 
> I think that the bounding box that should be updated is only for the case of a loading button, which is the one that has a secondary text, which causes the bug.
> 
> https://github.com/Telefonica/mistica-android/blob/432fb54363ad0252eabe4b79f482d56f67b8649b/library/src/main/java/com/telefonica/mistica/compose/button/Button.kt#L63
> 
> WDYT ?
> 
> Adit:
> 
> Also, I think the case with a loading text larger than standard text is not covered yet:
> 
> drawing

You are right! Changed.

The case when the loading text is longer than normal text will be implemented in the future. I have to check with team if we really need to cover this behaviour for now. Thinking about a possible solution, we need to calculate the width before painting the button. It maybe cause an overprograming.

jdelga (approved):
So cool! Thanks! ❤️

yceballost (approved):

yceballost (approved):

PR #361, opened 2024-05-31
ANDROID-13760 Avoid duplicated tags in Compose Tab element

pmartinbTEF:
Avoid duplicated tags in Compose Tab element

This error was caused by SubcomposeLayout. During measurement phase composables are created but they are not deleted when the measurement is finished.
The solution applied was to not set tags during measurement of composition. There are still duplicated elements but only those that are visible have a tag

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

- [x] Create Composable Tabs and assign a tag to each tab
- [x] Create a test in Espresso to find a tab by its tag
- [x] Check that only one element is found: composeRule.onAllNodesWithTag("tag").fetchSemanticsNode().size

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/tabs/Tabs.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/tabs/Tabs.kt
@@ -62,10 +62,10 @@ fun Tabs(
     val isTablet = LocalConfiguration.current.screenWidthDp > 768
     val maxTabWidth = 280.dp
 
-    val composableTabs: @Composable @UiComposable () -> Unit = @Composable {
+    val composableTabs: @Composable @UiComposable (setTestTag: Boolean) -> Unit = @Composable { setTestTag ->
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/telefonica/mistica/compose/tabs/Tabs.kt
```diff
@@ -62,10 +62,10 @@ fun Tabs(
     val isTablet = LocalConfiguration.current.screenWidthDp > 768
     val maxTabWidth = 280.dp
 
-    val composableTabs: @Composable @UiComposable () -> Unit = @Composable {
+    val composableTabs: @Composable @UiComposable (setTestTag: Boolean) -> Unit = @Composable { setTestTag ->
```
pmartinbTEF, line 65: Only set tag to composableTabsForPlacing

DevPabloGarcia (approved):
Thanks!

dpastor (approved):

PR #363, opened 2024-06-04
ANDROID-14761 Fix Image size of EmptyStateScreenView when it's IMAGE_SIZE_FULL_WIDTH

jeprubio:
In EmptyStateScreenView when the image type is IMAGE_SIZE_FULL_WIDTH it loads the image in full width but it doesn't use the ratio of 16:9 to assign the height. As it's also using ScaleType.CENTER_CROP as scale type it results in the image being cropped.

- In order to not having to add a GlobalLayoutListener in order to calculate the height async that would result in the image blinking the simplest way is to update the layout so that the image is inside a ConstraintLayout instead of a LinearLayout and that way the aspect ratio can be used to get the height.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

before:
mistica_before
after:
mistica_after

ratioCalculator

And when applied to vivo:
| update_screen_before     | update_screen_after |

- [x] Mistica App QR or [download link](https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/545)

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/emptystate/screen/EmptyStateScreenView.kt
+++ b/library/src/main/java/com/telefonica/mistica/emptystate/screen/EmptyStateScreenView.kt
@@ -15,6 +15,7 @@ import android.widget.TextView
 import androidx.annotation.DrawableRes
 import androidx.annotation.IntDef
 import androidx.appcompat.content.res.AppCompatResources
+import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.databinding.BindingMethod
 import androidx.databinding.BindingMethods
 import com.telefonica.mistica.R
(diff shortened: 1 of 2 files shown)


yceballost (approved):
nice :)

dpastor (approved):
Ok! 👍

pmartinbTEF (approved):
Nice! Did you try with vertical images? Did you try in api 24 device?

jeprubio:
> Nice! Did you try with vertical images? Did you try in api 24 device?

According to the specs it's only 16:9 images this type, I also checked the other types of images after doing the changes. I'm having trouble to create the api 24 emulator but I'll try again.

PR #366, opened 2024-06-06
ANDROID-14628 Not include release notes in the release notification

jeslat:
Not include release notes in the release notification. Just include the link to the release to avoid errors when it contains markdown.

I've checked with the GitHub API that `html_url` is the right field:
imagen


DevPabloGarcia (approved):
Nice catch

pmartinbTEF (approved):
Well done!

PR #372, opened 2024-07-11
ANDROID-14803 Update all the 24dp lateral margins to 16dp

pmartinbTEF:
Check that all sides margins of Mistica elements are 16dp (and not 24dp)

-Checked Mistica catalog using layout inspector to check sides margins
-Replaced 24dp with 16dp

Only found empty state xml implementation which used 24dp of side margins
I also found that link button of card_actions_view uses a negative margin. I think this should be reviewed in another task

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

- [x] Open Mistica catalog using layout inspector
- [x] Check that sides margins are 16dp

Before
drawing

After
drawing

### DIFF
--- a/library/src/main/res/layout/empty_state_screen_view.xml
+++ b/library/src/main/res/layout/empty_state_screen_view.xml
@@ -66,7 +66,7 @@
 					android:id="@+id/empty_state_screen_primary_button"
 					android:layout_width="wrap_content"
 					android:layout_height="wrap_content"
-					android:layout_marginEnd="24dp"
+					android:layout_marginEnd="@dimen/button_separation"
 					tools:text="Explore Marketplace"
 					/>
 
@@ -74,7 +74,7 @@
 					android:id="@+id/empty_state_screen_secondary_button"
 					android:layout_width="wrap_content"
 					android:layout_height="wrap_content"
-					android:layout_marginEnd="24dp"
+					android:layout_marginEnd="@dimen/button_separation"
 					style="@style/AppTheme.Button.SecondaryButton"
 					tools:text="Secondary Action"
 					/>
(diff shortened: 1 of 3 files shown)


library/src/main/res/layout/empty_state_screen_view.xml
```diff
@@ -69,15 +66,15 @@
 					android:id="@+id/empty_state_screen_primary_button"
 					android:layout_width="wrap_content"
 					android:layout_height="wrap_content"
-					android:layout_marginStart="24dp"
+					android:layout_marginEnd="24dp"
```
jeprubio, line 69: This is a bit confusing until you find it's meaning, perhaps it could be extracted to a dimension margin_between_buttons or someting like that.

Another perhaps better approach would be to leave them inside the ConstraintLayout and create a chain and set the paddings between elements.

jeprubio (approved):

library/src/main/res/layout/empty_state_screen_view.xml
```diff
@@ -7,9 +7,10 @@
 	<androidx.constraintlayout.widget.ConstraintLayout
 			android:layout_width="match_parent"
 			android:layout_height="wrap_content"
+			android:paddingStart="16dp"
 			android:paddingTop="64dp"
-			android:paddingEnd="24dp"
-			android:paddingBottom="24dp"
+			android:paddingEnd="16dp"
+			android:paddingBottom="16dp"
```
dpastor, line 13: padding bottom should be modified?

aweell: We should go with 16dp to match when the button is in the bottom on the screen the spacing the button fixed has.

aweell (approved):
LGTM

dpastor (approved):
👍

library/src/main/res/layout/empty_state_screen_view.xml
pmartinbTEF: I prefer the first option because it doesn't imply changing the code. Done!

jeprubio (approved):

PR #374, opened 2024-07-31
ANDROID-15026 make TextInput label param as nullable

DevPabloGarcia:
Currently "label" param is a not nullable string, this way, when we are developing if we pass empty value, the label is not shown, but the space is reserved and the value is not center vertically

* Make label param nullable

### ☑️ Checks
- [ ] I updated the documentation, including readmes and wikis. If this is a breaking change, tag the PR with "Breaking Change" label and remember to include breaking change migration guide in release notes where this version is released.
- [ ] Tested with dark mode.
- [ ] Tested with API 24.

- [ ] 🖼️ Screenshots/Videos

Captura de pantalla 2024-07-31 a las 10 20 52

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/input/EmailInput.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/input/EmailInput.kt
@@ -10,7 +10,7 @@ fun EmailInput(
     modifier: Modifier,
     value: String,
     onValueChange: (String) -> Unit,
-    label: String,
+    label: String?,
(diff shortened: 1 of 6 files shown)


yamal-alm (approved):
👍

yceballost (approved):
Use a input without label is considered a bad practice in terms of UX, but I'm agree to fix this case. Thanks!

haynlo (approved):
Nice! 🚀

PR #377, opened 2024-08-28
ANDROID-15075 ListRowItem Switch & CheckBox a11y compose

jeprubio:
Align ListRowItem accessibility for Compose according to new Figma definition: https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=4977-5442&t=iqE1tr998GXqKVZ4-4

- Creating ListRowItemWithSwitch and ListRowItemWithCheckBox ( b24f198a1603781be512ec2ab80da348465c90e3 )
- Letting developers to pass a contentDescription for badges ( 600bf16aced8749eeecd9efee930731977a315b6 )
- Adding info to the compose list README file.
- Scouting (unrelated task): Adding MisticaTheme to the PasswordInput so the previews in that file work again.

### ☑️ Checks
- [ ] Tested with dark mode.
- [ ] Tested with API 24.

- [ ] 🖼️ Screenshots/Videos
- [ ] ...

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
@@ -30,7 +30,9 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.semantics.heading
+import androidx.compose.ui.semantics.isTraversalGroup
 import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.traversalIndex
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.telefonica.mistica.R
@@ -40,7 +42,6 @@ import com.telefonica.mistica.compose.tag.Tag
 import com.telefonica.mistica.compose.theme.MisticaTheme
 import com.telefonica.mistica.compose.theme.brand.MovistarBrand
 
-@OptIn(ExperimentalMaterialApi::class)
 @Composable
 fun ListRowItem(
     modifier: Modifier = Modifier,
@@ -52,6 +53,7 @@ fun ListRowItem(
     backgroundType: BackgroundType = BackgroundType.TYPE_NORMAL,
     badge: String? = null,
     isBadgeVisible: Boolean = false,
+    isToggleable: Boolean = false,
     headline: Tag? = null,
     trailing: @Composable (() -> Unit)? = null,
     onClick: (() -> Unit)? = null,
@@ -68,6 +70,7 @@ fun ListRowItem(
         backgroundType = backgroundType,
         badge = badge,
         isBadgeVisible = isBadgeVisible,
+        isToggleable = isToggleable,
         headline = headline,
         trailing = trailing,
         onClick = onClick,
@@ -76,7 +79,6 @@ fun ListRowItem(
     )
 }
 
-@OptIn(ExperimentalMaterialApi::class)
 @Composable
 @Deprecated(replaceWith = ReplaceWith("ListRowItem"), message = "Use new ListRowItem with ListRowIcon param instead")
 fun ListRowItem(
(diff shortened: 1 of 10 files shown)


dpastor (approved):
Good work!

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -125,6 +127,7 @@ private fun ListRowItemImp(
     backgroundType: BackgroundType = BackgroundType.TYPE_NORMAL,
     badge: String? = null,
     isBadgeVisible: Boolean = false,
+    isToggleable: Boolean = false,
     headline: Tag? = null,
     trailing: @Composable (() -> Unit)? = null,
     onClick: (() -> Unit)? = null,
```
haynlo, line 133: I think we should also handle onClick param when it's not null in order to create a custom ContentDescription since Talkback will announce by defult the natural order of elements: [Headline + Title + Subtitle + Description + etc.].

According to Figma definition [[link](https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=4615-10711&t=19Q7PVobSdYc1MwL-4)] it should be different, one approach could be this one:
```kotlin
if (onClick != null) {
   Modifier.semantics {
      contentDescription = /* build here custom order */
   }   
}
```

This custom contentDescription has been also built in the XML side for clickable and non-clickable elements due to XLM limitations, but you can check it for reference o even reuse its utility classes: https://github.com/Telefonica/mistica-android/blob/main/library/src/main/java/com/telefonica/mistica/list/ListRowView.kt#L185

haynlo (changes requested):

library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
jeprubio: Have you tried or watched the videos before? I don't think it does follow natural order in any scenario.

jeprubio: I see what you mean, sorry, that was broken when fixing the ordering of the non clickable rows.
Fixed, without having to set a content description.

https://github.com/user-attachments/assets/9b6236f8-eda1-40c5-967f-91592adad1ac

haynlo (approved):
Great job! 👏🏼

aweell:
### Interactive lists in compose

According to spec, the interactive lists should be like one element to the talkback focus, in classic this is working as expected, but in compose seems that no matter the list is informative or interactive (navigation) the elements are focused individually.

Video of the behaviour:

https://github.com/user-attachments/assets/a7e6a8ad-6bb2-4944-80ed-eed41dc54afc

(Some texts are not announced, but this is an issue with my phone not able to handle talkback and screen recording at the same time

The version where I'm seeing this issue is: https://install.appcenter.ms/orgs/tuenti-organization/apps/mistica-catalog/distribution_groups/public/releases/592

aweell:
https://github.com/Telefonica/mistica-android/pull/377#issuecomment-2328421664

Resolved offline, I was assuming that all list w/chevron were interactive, but the cases tested didn't have onClick.

The behaviour of the interactive list in compose works as expected

aweell:
## Badges behaviour in list with onClick

The badge should no have an independent focus of the list, currently it seems that:

* Classic: the badge is included in the concatenated content (expected behaviour)
* Compose: the badge is not included and has an independent focus after the list has been focused

## Read badges content

I see value in allow to provide a contentDescription for the badge component i would keep that change in the badge for other uses.

In the list component we decided to remove the content of the badges of the concatenated content and make them in all cases invisible to screen readers (so the should not be focusable content even if the list is informative).

Why this? 

* If a content description is not provided, the numeric content of the badge is not giving the user a lot of information about what that number means, that depends of the context and we can't control the context.
* If a content description is provided, we cannot anticipated in which position the badge should be ordered when concatenating the text.

The option we concluded that was the best experience for users is to delegate that responsibility in the product implementation phase, where the context can be controlled, allowing to provide a custom accesibilityLabel to the whole list where this information can be included in the right way.

PR #382, opened 2024-09-10
Use specific version when notifying Teams

jeslat:
Use specific version when notifying Teams to avoid unexpected changes.


haynlo (approved):
👍🏼

PR #387, opened 2024-09-19
ANDROID-15206 Set icon as null when there is no icon

jeslat:
Set icon as null when there is no icon. Due to this, the left padding was wrong in the compose list row item because we were adding an extra space because the icon was not null.

### ☑️ Checks

| drawing      | drawing |

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
@@ -60,7 +60,7 @@ fun ListRowItem(
 ) {
     ListRowItemImp(
         modifier = modifier,
-        icon = { listRowIcon?.Draw() },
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/telefonica/mistica/compose/list/ListRowItem.kt
```diff
@@ -293,6 +293,10 @@ fun ListRowItemPreview() {
     MisticaTheme(brand = MovistarBrand) {
         val checkedState = remember { mutableStateOf(true) }
         Column {
+            ListRowItem(
```
jeslat, line 296: Just updating the preview to add this case

juangardi21 (approved):

dpastor (approved):
👍

yceballost (approved):
👌

PR #395, opened 2024-11-06
Fix acceptance errors of v15.0.1

pmartinbTEF:
Fix acceptance errors of v15.0.1. PerformClick call was causing double calls to progressButton. Also if you scroll starting from the button, lifting your finger activates the button.

Deleted performClick call and added SuppressLint annotation

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

Tested [in this PR](https://github.com/Telefonica/android-messenger/actions/runs/11703406958/job/32593677318?pr=2452) with this [snapshot (15.0.1.2-SNAPSHOT)](https://github.com/Telefonica/mistica-android/actions/runs/11702882702/job/32592000326)

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/button/ProgressButton.kt
+++ b/library/src/main/java/com/telefonica/mistica/button/ProgressButton.kt
@@ -191,7 +191,6 @@ class ProgressButton : FrameLayout {
         }
     }
 
-    @SuppressLint("ClickableViewAccessibility")
     private fun setButtonBackground() {
         buttonBackground.apply {
             id = NO_ID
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/telefonica/mistica/button/ProgressButton.kt
```diff
@@ -191,6 +191,7 @@ class ProgressButton : FrameLayout {
         }
     }
 
+    @SuppressLint("ClickableViewAccessibility")
```
pmartinbTEF, line 194: Talkback working properly

jeprubio (approved):

yamal-alm (approved):

jdelga (approved):

library/src/main/java/com/telefonica/mistica/button/ProgressButton.kt
haynlo: Also SwitchAccess has been verified ✅

PR #396, opened 2024-11-08
ANDROID-15325 Compose Input test ids

yamal-alm:
Test IDs for input component (only compose by now)

I added the following IDs:
- TextField
- PasswordField
- SearchField
- PhoneNumberField
- EmailField
- DecimalField
- label
- helperText
- endHelperText
- errorText
- startIcon
- endIcon

Following IDs weren't added because associated input fields or features are not implemented yet:
- CreditCardexpirationField
- CreditCardNumberField
- CvvField
- DateField
- DateTimeField
- IbanField
- IntegerField
- MonthField
- PinField
- placeholder

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
@@ -7,6 +7,7 @@ object TextInputTestTags {
     const val PHONE_NUMBER_INPUT = "PhoneNumberField"
     const val EMAIL_INPUT = "EmailField"
     const val DECIMAL_INPUT = "DecimalField"
+    const val LABEL = "label"
     const val HELPER_TEXT = "helperText"
     const val END_HELPER_TEXT = "endHelperText"
     const val ERROR_TEXT = "errorText"
--- a/library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
@@ -1,7 +1,18 @@
 package com.telefonica.mistica.compose.input
 
 object TextInputTestTags {
-    const val TEXT_INPUT = "text_input"
+    const val TEXT_INPUT = "TextField"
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
```diff
@@ -1,7 +1,19 @@
 package com.telefonica.mistica.compose.input
 
 object TextInputTestTags {
-    const val TEXT_INPUT = "text_input"
+    const val TEXT_INPUT = "TextField"
```
yamal-alm, line 4: because of this, the release including these changes will contain breaking changes

yceballost (approved):

library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
pmartinbTEF: Why some tags start with uppercase and others don't?

pmartinbTEF (commented):
Nice! just one question

library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
yamal-alm: that's the definition from Figma

pmartinbTEF (approved):

library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
DevPabloGarcia: I think we should check with QA team about this definition. I think we need to create a common rule to naming

DevPabloGarcia (approved):
Yeah !

library/src/main/java/com/telefonica/mistica/compose/input/TextInputTestTags.kt
yamal-alm: isn't this the final definition?

PR #398, opened 2024-11-21
ANDROID-15405 -- PosterCard component

dhonti-axpe:
First approach about **PosterCard** component creation according to current [Card Spec section](https://www.figma.com/design/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?node-id=2085-6956&node-type=frame&t=gMC4TdFdciN5iHiz-0).

Links references:
* [Brand factory documentation](https://brandfactory.telefonica.com/d/iSp7b1DkYygv/n-a#/components/cards/poster-card)
* [Playground](https://mistica-web.vercel.app/?path=/story/components-cards-postercard--default)

* First of all, a `PosterCard.kt` file exposes the main component through is provided corresponding implementation.
* Corresponding section in library catalog README.md
* This component is divided in 3 sections:
    * **TopSection**: Possibility to add a maximum of 2 TopAction actionable rounded (or not) icons
    * **MainSection**
         * Tag
         * Pretitle
         * Title
         * Subtitle
         * Description
    * **CustomSection**: A custom composable that will be integrated at bottom of current component

> [!WARNING]  
> This implementation does not include the **background video**

### ☑️ Checks
- [X] I updated the documentation, including readmes and wikis. If this is a breaking change, tag the PR with "Breaking Change" label and remember to include breaking change migration guide in release notes where this version is released.
- [X] Tested with dark mode.
- [X] Tested with API 24.
- [X] Sync done with iOS team for this feature to ensure alignment, if applies.
- [X] Accessibility considerations.

- [X] 🖼️ Screenshots/Videos

| Main functionality   | A11 integrated  |
| ------------------ | ---------------- |
| <video  src="https://github.com/user-attachments/assets/171ed62b-7014-408f-b995-bfcbffea7f9d" alt="PosterCard main functionality" width="200" /> | <video  src="https://github.com/user-attachments/assets/181f1246-3dba-4053-8f67-b8e4a4743074" alt="PosterCard main functionality" width="200" /> |

- [X] Mistica App QR or download link
- [X] Reviewed by Mistica design team


library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
```diff
@@ -0,0 +1,94 @@
...
+    description: String? = null,
+    firstTopAction: TopActionData? = null,
+    secondTopAction: TopActionData? = null,
+    onClickAction: (() -> Unit)? = null,
+    customContent: (@Composable () -> Unit)? = null,
+) {
+    BoxWithConstraints(modifier = modifier) {
+        androidx.compose.material.Card(
+            elevation = 0.dp,
+            backgroundColor = Color.Red,
```
haynlo, line 39: Why this red color?

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
```diff
@@ -0,0 +1,94 @@
...
+            shape = RoundedCornerShape(16.dp),
```
haynlo, line 40: According to Figma spec, should be "container" value instead of fixed 16.dp right?
https://www.figma.com/design/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?node-id=2085-7952&t=7A6BCRoDU8omYQoY-4
```suggestion
            shape = RoundedCornerShape(MisticaTheme.radius.containerBorderRadius),
```

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
```diff
@@ -0,0 +1,94 @@
...
+            border = BorderStroke(width = 1.dp, color = MisticaTheme.colors.border),
```
haynlo, line 41: I'm not sure if the card should have border stroke (In Figma's definition it seems that it doesn't have it). Can you confirm that please?
https://www.figma.com/design/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?node-id=2157-7504&t=7A6BCRoDU8omYQoY-4

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
```diff
@@ -0,0 +1,94 @@
...
+            modifier = Modifier
+                .width(maxWidth)
+                .let { modifierValue ->
```
haynlo, line 44: Minor: if you want to add conditioned configuration you can better use `then()` method like:
``` kotlin
Modifier.then(onClickAction?.let {
        Modifier.clickable { it.invoke() }
    } ?: Modifier)
    
```

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
```diff
@@ -0,0 +1,79 @@
...
+            .fillMaxWidth()
+            .padding(top = 16.dp, start = 16.dp, end = 16.dp, bottom = 8.dp),
+        horizontalArrangement = Arrangement.End,
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        firstTopAction?.let {
+            TopAction(topActionData = it)
+        }
+        secondTopAction?.let {
+            Spacer(modifier = Modifier.width(16.dp))
```
haynlo, line 40: Mmm.. with this logic there's a case where user can define a secondTopAction without a firstTopAction. In that case, we are showing the secondTopAction as a the first one but with this  unnecessary Space.

Maybe we should adjust this logic in order to admit only second action if there's a first one, or at least show a warning log. Wdyt?

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
```diff
@@ -0,0 +1,79 @@
...
+internal fun TopAction(topActionData: TopActionData) {
+    with(topActionData) {
+        Box(
+            modifier = Modifier
+                .testTag(topActionData.testTag.orEmpty())
+                .size(40.dp)
+                .clip(CircleShape)
+                .background(color = backgroundColor)
+                .wrapContentSize(align = Alignment.Center)
+                .let { modifierValue ->
```
haynlo, line 56: Same as mentioned before, you can use `then` instead

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+}
+
+@Composable
+internal fun TopAction(topActionData: TopActionData) {
+    with(topActionData) {
+        Box(
+            modifier = Modifier
+                .testTag(topActionData.testTag.orEmpty())
```
haynlo, line 51: Since you are using `with` scope function you can use:
```suggestion
                .testTag(testTag.orEmpty())
```

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
```diff
@@ -0,0 +1,94 @@
...
+enum class PosterCardAspectRatio(val ratio: Float) {
+    AR_AUTO(ratio = Float.NaN),
+    AR_1_1(ratio = 1f),
+    AR_7_10(ratio = 0.7f),
+    AR_9_10(ratio = 0.9f),
+    AR_16_9(16 / 9f)
+}
+
+sealed class PosterCardBackgroundType(open var inverseDisplay: Boolean) {
+    data class Image(val imageResource: Int, val contentDescription: String = "") : PosterCardBackgroundType(inverseDisplay = true)
```
haynlo, line 91:
```suggestion
    data class Image(val imageResource: Int, val contentDescription: String? = null) : PosterCardBackgroundType(inverseDisplay = true)
```
Same definition as you have done here: https://github.com/Telefonica/mistica-android/pull/398/files#diff-dd2b9dff03ae4ec8918409185caf888d1146bb8ba1b837173e163a58d6f319b7R76 :)

haynlo (commented):
I left you some comments, but most of them are just very minor suggestions and others are doubts between you implementation and the figma's definition.

The latter is the most important thing to clarify, thanks!

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
dhonti-axpe: At first, like is a Card type, I assumed that will take a border in the visual representation. However checking the spec there is not any value corresponding to current border. Fixed in next!

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
dhonti-axpe: Instead of use that, I think that is possible to use better something like:
```
.clickable(enabled = onClick != null) {
    onClick?.invoke()
}
```

WDYT?
(Applied in this commit) --> https://github.com/Telefonica/mistica-android/pull/398/commits/91ca8546f3096629afe6e6c50d3c82ae7a6b7377

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: Same response like here --> https://github.com/Telefonica/mistica-android/pull/398#discussion_r1857878790

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: Applied here --> https://github.com/Telefonica/mistica-android/pull/398/commits/91ca8546f3096629afe6e6c50d3c82ae7a6b7377

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
dhonti-axpe: So, in order to confirm, Who would I have to ask that, @haynlo ?

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: I provided in this commit --> https://github.com/Telefonica/mistica-android/pull/398/commits/3586a3619766976126a880f192d1e7a0d9889750
A new approach receiving a `TopActionData` list. In case list are null or the number of elements are upper than 2, list won't be displayed. I need to update the corresponding README.md file with this constraint.

WDYT?

yceballost (changes requested):

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardMainContent.kt
```diff
@@ -0,0 +1,132 @@
...
+    blurRadius = 15f
+)
+
+private fun buildTextBackgroundGradientBrush(backgroundType: PosterCardBackgroundType): Brush =
+    if (backgroundType.inverseDisplay) {
+        Brush.verticalGradient(
+            colorStops = arrayOf(
+                0.0f to Color.Black.copy(alpha = 0f),
+                0.3f to Color.Black.copy(alpha = 0.4f),
+                1.0f to Color.Black.copy(alpha = 0.7f)
```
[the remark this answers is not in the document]
dhonti-axpe: Thanks! Fixed in this commit --> https://github.com/Telefonica/mistica-android/pull/398/commits/faf034f9b6957d29d7f73a0168d2e01dc649c81f

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
haynlo: You can ask in the Mistica public Teams channel, por example: https://teams.microsoft.com/l/channel/19%3A68854aea147e4787803aff0e380e1208%40thread.tacv2/Help?groupId=e265fe99-929f-45d1-8154-699649674a40&tenantId=9744600e-3e04-492e-baa1-25ec245c6f10

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
haynlo: I didn't know the `enabled` param, looks good!

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
haynlo: Allowing the user to add a list of n items is not entirely correct, since if they add 3 or more items, none of them will be displayed.

I still prefer the previous approach, but we can use a Pair instead to make code more readable:
``` kotlin
internal fun PosterCardTopActions(modifier: Modifier = Modifier, topActionsList: Pair<TopActionData?, TopActionData?>) {
    Row( /* ... */ ) {
        topActions.first?.let {
            // ...
        }

        if (topActions.showSpaceBetween()) Spacer(modifier = Modifer.with(16.dp))

        topActions.second?.let {
            // ...
        }
    }
}

private fun Pair<String?, String?>.showSpaceBetween() = first != null && second != null   
```

If you din't want to use Pair values, you can use specific params like in your original approach but I think it's better to provide specific params instead of a generic list, Wdyt?

haynlo (commented):
I don't see any accessibility implementation.. should it be implemented here o will be implemented in a future task? https://www.figma.com/design/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?node-id=8552-8979&t=XBKHFm6OGwcgaRQj-4

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardPreviews.kt
```diff
@@ -0,0 +1,115 @@
...
+                TopActionData(
+                    iconRes = R.drawable.icn_visibility,
+                    backgroundColor = MisticaTheme.colors.backgroundAlternative,
+                    iconTint = MisticaTheme.colors.brand,
+                ),
+                TopActionData(
+                    iconRes = R.drawable.ic_close_regular,
+                    backgroundColor = MisticaTheme.colors.backgroundAlternative,
+                    iconTint = MisticaTheme.colors.brand,
+                )
```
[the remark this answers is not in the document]
dhonti-axpe: Included in this commit --> https://github.com/Telefonica/mistica-android/pull/398/commits/ef9ca70dbd89b026ba3901740da2e7e1ebde242e
Change the background with a initial opacity. When user press the button a clickable effect is shown natively in Android.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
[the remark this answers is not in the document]
dhonti-axpe: Previously due to the topActions does not receive any `onClick` event in Catalog does not show any clickable action. Now by default, a empty clickable actions is provided to sample that interaction. Also in this commit --> https://github.com/Telefonica/mistica-android/commit/ef9ca70dbd89b026ba3901740da2e7e1ebde242e

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: Personally I prefer recovering previous implementation  (defining a `firstTopAction` and a `secondTopAction` as independent params), since if you provide a `Pair` type param, you need to provide both (`first` and `second`) params always, regardless you don't want to provide one of them as NULL value.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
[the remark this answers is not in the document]
dhonti-axpe: I changed to previous version in this commit separating `firstAction` and `secondAction` as individual params --> https://github.com/Telefonica/mistica-android/pull/398/commits/f1f1d21c7e7760b07d35c19df2b70df84537f464

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: I changed to previous version in this commit (https://github.com/Telefonica/mistica-android/pull/398/commits/f1f1d21c7e7760b07d35c19df2b70df84537f464) separating `firstAction` and `secondAction` as individual params.

library/src/main/res/drawable-nodpi/sample_background.jpg
dpastor, line 1: This file is inside the library source code, can we put it in the catalog? (This increases the library size in 29.3kb, which is not needed here, right?)

dpastor (approved):
👍

haynlo (commented):
Waiting for a11y implementation for final review.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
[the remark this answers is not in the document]
yceballost: I can't see the 48x48 tap area haha, but I will trust you! 
image

yceballost (approved):
nice job Ɛ>

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
```diff
@@ -0,0 +1,100 @@
...
+}
+
+@Composable
+internal fun TopAction(
+    modifier: Modifier = Modifier,
+    topActionData: TopActionData
+) {
+    with(topActionData) {
+        Box(
+            modifier = Modifier
```
jeprubio, line 62: I think this should be modifier = modifier and not in the then at the end. If you want to add some external paddings, alignment to a box... to the TopAction it won't allow you this way it's done in here if I'm not wrong.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
```diff
@@ -0,0 +1,100 @@
...
+) {
+    Row(
+        modifier = modifier
+            .fillMaxWidth()
+            .padding(16.dp),
+        horizontalArrangement = Arrangement.End,
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        firstTopAction?.let {
+            TopAction(topActionData = it, modifier = Modifier.semantics { traversalIndex = 7f })
```
jeprubio, line 44: Does traversalIndex work without defining a traversalGroup? My experience was that it doesn't work, idk.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardBackground.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.compose.ui.res.painterResource
+
+@Composable
+internal fun PosterCardBackground(
+    modifier: Modifier = Modifier,
+    backgroundType: PosterCardBackgroundType,
+    content: @Composable BoxScope.() -> Unit
+) {
+    Box(
+        modifier = Modifier
```
jeprubio, line 21: Here also I think the parameter should be the first modifier to apply and not the last one.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: I forgot remove that reference to semantics in this composable. Change in next commit

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: I forgot remove that reference to semantics in this composable. Change in next commit

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
dhonti-axpe: You're so right!

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCard.kt
```diff
@@ -0,0 +1,98 @@
...
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.zIndex
+import com.telefonica.mistica.compose.tag.Tag
+import com.telefonica.mistica.compose.theme.MisticaTheme
+
+@Composable
+fun PosterCard(
+    modifier: Modifier = Modifier,
```
jeprubio, line 25: https://blog.simprasuite.com/jetpack-compose-respect-the-contract-of-modifiers-ecbbe8ce03db#:~:text=The%20modifier%20should%20be%20the,other%20parameters%20with%20default%20values.
> The modifier should be the first optional parameter in the parameter list; after all required parameters (except for trailing lambda parameters like card content) but before any other parameters with default values.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardBackground.kt
jeprubio: I didn't ask for removal of the modifier parameter, just applying it well following the best practices defined in the compose api guidelines:
https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter

> Element functions MUST accept a parameter of type Modifier. This parameter MUST be named “modifier” and MUST appear as the first optional parameter in the element function's parameter list. Element functions MUST NOT accept multiple Modifier parameters.

library/src/main/java/com/telefonica/mistica/compose/card/postercard/PosterCardTopActions.kt
jeprubio: I didn't ask for removal of the modifier parameter, just applying it well following the best practices defined in the compose api guidelines:
https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter

> Element functions MUST accept a parameter of type Modifier. This parameter MUST be named “modifier” and MUST appear as the first optional parameter in the element function's parameter list. Element functions MUST NOT accept multiple Modifier parameters.

jeprubio (approved):
👍

PR #400, opened 2024-12-19
ANDROID-14792 Support Checkbox with Links for Compose implementation

haynlo:
### :tickets: Jira ticket
[ANDROID-14792](https://jira.tid.es/browse/ANDROID-14792)

### :goal_net: What's the goal?
Support Checkbox with Links for Compose implementation

**⚠️ Don't panic!** The actual changes affect only less than 10 files. All other changes are related to the baseline ScreenShot due to the migration of the RippleTheme.

### :construction: How do we do it?
* Update Compose UI version
* Fix problemas related to Compose UI upgrade
  * RippleTheme is now deprecated. It has been migrated.
* Apply Role.Checkbox to Row wrapper to improve component usability.
* Change TextWithLinks Composable in order to support new `spannableString` `withLink` method

### ☑️ Checks
- [x] I updated the documentation, including readmes and wikis. If this is a breaking change, tag the PR with "Breaking Change" label and remember to include breaking change migration guide in release notes where this version is released.
- [x] Tested with dark mode.
- [x] Tested with API 24.
- [x] Sync done with iOS team for this feature to ensure alignment, if applies.
- [x] Accessibility considerations.

#### Ripple effect still being the same after the migration 
| Ripple Before        | Ripple After           |
| ------------- |-------------|
| <video src="https://github.com/user-attachments/assets/acc0a829-7457-4717-8874-2b9c8a0c71f5" alt="video showing ripple before the migration" width="200"/>      | <video src="https://github.com/user-attachments/assets/61ce7cbc-cd5c-45bd-9917-2ef8a607380f" alt="video showing ripple after the migration" width="200"/> |

#### Demo of main change
| CheckBox with links in Compose        | 
| ------------- |
| <video src="https://github.com/user-attachments/assets/aeb1a9f8-ff70-4810-ac94-1e308f225b5b" alt="video showing ripple before the migration" width="200"/>      |

PR #405, opened 2025-01-20
ANDROID-15643 Add test tag to Checkbox

yamal-alm:
`CheckBoxInput` implementation was changed here https://github.com/Telefonica/mistica-android/pull/400 and now the whole `Row` that wraps the `Checkbox` and `TextWithLinks` composables is toggable.

That is more correct in terms of a11y but unexpectedly it has broken some tests in client apps when trying to click on a checkbox with links. The reason is because trying to do some click on a `CheckBoxInput` with a given `testTag` it ends up in a click on the middle of the wrapping `Row` previously mentioned and it may open any link if present.

To prevent that we propose to add some inner testTag to the single Checkbox.

Adding some test tag to the `Checkbox` composable from material user under the hood by `CheckBoxInput`

### ☑️ Checks

### DIFF
--- a/library/src/main/java/com/telefonica/mistica/compose/input/CheckBoxInput.kt
+++ b/library/src/main/java/com/telefonica/mistica/compose/input/CheckBoxInput.kt
@@ -9,6 +9,7 @@ import androidx.compose.material.Checkbox
 import androidx.compose.material.CheckboxDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
@@ -35,6 +36,7 @@ fun CheckBoxInput(
             role = Role.Checkbox
         )) {
             Checkbox(
+                modifier = Modifier.testTag(CheckBoxInputTestTags.CHECKBOX),
                 checked = checked,
                 onCheckedChange = null,
                 enabled = enabled,
@@ -97,4 +99,8 @@ fun PreviewDisabledCheckBoxInput() {
         text = "Disabled checkbox",
         enabled = false,
     )
+}
+
+object CheckBoxInputTestTags {
+    const val CHECKBOX = "checkbox_input_checkbox"
 }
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/telefonica/mistica/compose/input/CheckBoxInput.kt
```diff
@@ -98,3 +100,7 @@ fun PreviewDisabledCheckBoxInput() {
         enabled = false,
     )
 }
+
+object CheckBoxInputTestTags {
+    const val CHECKBOX = "checkbox_input_checkbox"
```
yamal-alm, line 105: I'm open to naming suggestions

haynlo: Maybe we can add the CheckBox.testTag value as a parameter in order to give more flexibility in each implementation for testing:
```kotlin
@Composable
fun CheckBoxInput(
   ...
   checkboxTestTag: String = CheckBoxInputTestTags.DEFAULT,
) { }
```
wdyt?

yamal-alm: I think that is not necessary because we can do things like following:
```kotlin
composeTestRule.waitForNodeDisplayed {
    it.onNode(
        hasTestTag(CheckBoxInputTestTags.DEFAULT).and(hasAnyAncestor(hasTestTag(customTag)))
    )
}
```

Where `customTag` is the tag that we can set to the `CheckBoxInput` composable through its `Modifier`

haynlo: You are totally right. In that case let's keep it simpler as you have defined it 🙂

haynlo (approved):
Thanks for the update!
PS: Component a11y still works perfectly ^^

library/src/main/java/com/telefonica/mistica/compose/input/CheckBoxInput.kt
DevPabloGarcia: Maybe we can name it just checkbox_input or checkbox (keeping the nomenclature of figma ids)

example: https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=4615-10677&t=ukaY9yT3l8UMeZ2i-0

DevPabloGarcia (commented):
Naming suggestion

library/src/main/java/com/telefonica/mistica/compose/input/CheckBoxInput.kt
yamal-alm: sure! f16c877652205fd01fd1628a9029e662b5d3137d

DevPabloGarcia (approved):
😄

PR #412, opened 2025-04-04
ANDROID-15982 Fix feedback Accessibility

jeprubio:
In the feedback component in mística the image, title, subtitle are read grouped by talkback.
Also the title should be set as a heading.
This PR is in order to fix that.

- Removed focusable on the LinearLayout that made the elements be grouped.
- Set the title as heading.

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.
- [x] Sync done with iOS team for this feature to ensure alignment, if applies. <-- iOS has already these elements separate

**_Turn up the volume on the videos_**

| <video src="https://github.com/user-attachments/assets/eb9127f0-94c8-412f-a117-74d69252566c" alt="Before" width="200"/>      | <video src="https://github.com/user-attachments/assets/4ae8c36a-aadc-43f2-823a-5bbbab50e148" alt="After" width="200"/> |


yceballost (approved):
nice!

PR #413, opened 2025-04-07
ANDROID-15018 Callout a11y

dagonco:
- Make the title view configurable so that talkback treats it as a header or not. By default, **it will always** be a header.
- Make the close/dismiss button the last node in the accessibility hierarchy.

* Expose a method to configure "heading"
* Set close/dismiss button to be the last in the hierarchy
* Fixed XML view close/dismiss button touch zone.

### ☑️ Checks
- [X] I updated the documentation, including readmes and wikis. If this is a breaking change, tag the PR with "Breaking Change" label and remember to include breaking change migration guide in release notes where this version is released.
- [X] Tested with dark mode.
- [X] Tested with API 24.
- [X] Accessibility considerations.

- [X] 🖼️ Screenshots/Videos
- [X] Mistica App QR or download link
- [X] Reviewed by Mistica design team

| Focus Before        | Focus After           |
| drawing | drawing      |

| A11y XML        | A11y Compose           |
| <video src="https://github.com/user-attachments/assets/e9ce73f0-576c-4558-864c-ae70b0bdf40c" alt="video" width="200"/> | <video src="https://github.com/user-attachments/assets/d3995d9a-d6f9-476b-b26c-a27b02998d30" alt="video" width="200"/>      |

### DIFF
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Callouts.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Callouts.kt
@@ -148,7 +148,7 @@ fun Callouts() {
                         .fillMaxWidth()
                         .padding(16.dp),
                     title = title.takeIf { it.isNotBlank() },
-                    setTitleAsHeader = true,
+                    setTitleAsHeading = true,
                     description = description.takeIf { it.isNotBlank() },
                     buttonConfig = buttonConfig,
                     imageConfig = iconType,
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Callouts.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Callouts.kt
@@ -148,6 +148,7 @@ fun Callouts() {
                         .fillMaxWidth()
                         .padding(16.dp),
                     title = title.takeIf { it.isNotBlank() },
+                    setTitleAsHeader = true,
                     description = description.takeIf { it.isNotBlank() },
                     buttonConfig = buttonConfig,
                     imageConfig = iconType,
--- a/library/src/main/java/com/telefonica/mistica/callout/CalloutView.kt
+++ b/library/src/main/java/com/telefonica/mistica/callout/CalloutView.kt
@@ -33,7 +33,7 @@ import com.telefonica.mistica.util.getThemeColor
     BindingMethod(
         type = CalloutView::class,
         attribute = "calloutTitleAsHeader",
-        method = "setTitleAsHeader",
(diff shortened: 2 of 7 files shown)


library/src/main/java/com/telefonica/mistica/callout/CalloutView.kt
```diff
@@ -298,53 +311,58 @@ class CalloutView @JvmOverloads constructor(
...
 
     fun setButtonsConfig(@ButtonsConfig buttonsConfig: Int) {
-        if (buttonsConfig == BUTTONS_CONFIG_NONE) {
-            buttonsContainer.visibility = GONE
-            return
-        }
         buttonsContainer.visibility = VISIBLE
 
         when (buttonsConfig) {
+            BUTTONS_CONFIG_NONE -> {
```
dagonco, line 317: We had this logic out of the `when`... It seems not to make sense.

library/src/main/res/layout/callout_view.xml
```diff
@@ -9,15 +9,16 @@
...
 
-	<ImageView
+	<ImageButton
 			android:id="@+id/callout_close_button"
 			app:layout_constraintTop_toTopOf="parent"
 			app:layout_constraintRight_toRightOf="parent"
-			android:translationX="4dp"
-			android:layout_width="24dp"
-			android:layout_height="24dp"
+			android:background="@android:color/transparent"
```
dagonco, line 16: Check screenshots in PR's description. I nail same position.

library/src/main/java/com/telefonica/mistica/callout/CalloutView.kt
haynlo: I think this was done to avoid showing the Container and immediately after hiding it and its possible issues with that (Probably none, but I'm not sure).

haynlo (approved):
Nice! 🚀

catalog/src/main/java/com/telefonica/mistica/catalog/ui/compose/components/Callouts.kt
```diff
@@ -148,6 +148,7 @@ fun Callouts() {
                         .fillMaxWidth()
                         .padding(16.dp),
                     title = title.takeIf { it.isNotBlank() },
+                    setTitleAsHeader = true,
```
aweell, line 151: Why header instead of heading?

dagonco: Well, it's just our code naming, it does not affect to the feature. Didn't know that Header and Heading could have differences. But I'll rename it

jeslat (approved):
👍

aweell (approved):

PR #419, opened 2025-04-30
ANDROID-16118 Escape release notes to avoid problems with multiline

jeslat:
Escape release notes to avoid problems with multiline.

* Add a new step to escape the release notes using `jq` (thanks copilot) to avoid problems when they are multiline and are added to the JSON.

I've done some tests setting the release notes text as an environment variable and worked ok:

image


dagonco (approved):
Thanks Yisas

jdelga (approved):

PR #438, opened 2025-07-04
ANDROID-15258 List accessibility update for XML implementation

haynlo:
### :tickets: Jira ticket
[ANDROID-15258](https://jira.tid.es/browse/ANDROID-15258)

ℹ️ **Only for XML implementation**. The Compose implementation will be added in this other task: https://jira.tid.es/browse/ANDROID-15075

- Allow to provide a custom label in informative rows to be read by screen readers in a single focus, instead reading all elements one by one.
- Hide always badges to Screen readers no matter if the list is informative or interactive

For more info, check out this Figma documentation: https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=4615-10678&t=eRMRYwPaobn7VxjU-4

* Remove badgeDescription implementation.
* Managing the coexistence both contentDescription:
   * Default: handled by the component itself
   * Custom: managed by the implementation
* Create unitTests for ensure the proper behavior with the new contentDescription logic.
* Add readme to explain the users how to handle the new custom contentDescription

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.
  ℹ️ Since iOS isn't currently handling this task, I'll add a comment to the iOS task to inform them about the Android implementation.

https://github.com/user-attachments/assets/194cb411-d749-4dba-9db1-c939fd0a8853

### DIFF
--- a/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
+++ b/catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
@@ -102,7 +102,8 @@ class ListsCatalogFragment : Fragment() {
                 withLongTitle: Boolean = false,
                 withTitleMaxLines: Int? = null,
                 withDescriptionMaxLines: Int? = null,
-                withDimensions: ImageDimensions? = null
+                withDimensions: ImageDimensions? = null,
+                withCustomContentDescription: Boolean = false,
             ): (ListRowView) -> Unit = { view ->
                 view.configureView(
                     withAction = withAction,
@@ -120,13 +121,15 @@ class ListsCatalogFragment : Fragment() {
                     withLongTitle = withLongTitle,
                     withTitleMaxLines = withTitleMaxLines,
                     withDescriptionMaxLines = withDescriptionMaxLines,
-                    withDimensions = withDimensions
+                    withDimensions = withDimensions,
+                    withCustomContentDescription = withCustomContentDescription,
                 )
             }
 
             return listOf(
                 configure(withTitleHeading = true),
                 configure(withAction = true),
+                configure(withCustomContentDescription = true),
                 configure(withAction = true, withBadge = true),
                 configure(withAction = true, withBadgeNumeric = 1),
                 configure(withLongDescription = false, withTitleHeading = true),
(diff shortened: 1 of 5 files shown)


catalog/src/main/java/com/telefonica/mistica/catalog/ui/classic/components/ListsCatalogFragment.kt
```diff
@@ -165,14 +166,13 @@ class ListsCatalogFragment : Fragment() {
...
-
         @SuppressLint("SetTextI18n")
         @Suppress("CyclomaticComplexMethod")
         private fun ListRowView.configureView(
             withLongTitle: Boolean = false,
             withTitleMaxLines: Int? = null,
-            withTitleHeading: Boolean? = false,
-            withLongDescription: Boolean? = null,
+            withTitleHeading: Boolean = false,
+            withLongDescription: Boolean = true,
```
haynlo, line 175: Minor improvement, not related with the main change.

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -181,6 +180,7 @@ open class ListRowView @JvmOverloads constructor(
     private var cachedDefaultBackgroundType: Int = BackgroundType.TYPE_NORMAL
 
     private var isViewInitialized = false
+    private var hasCustomContentDescription = false
```
haynlo, line 183: Flag needed to check if current contentDescription is provided by the implementation (custom) or by the component itself (default)

library/src/main/java/com/telefonica/mistica/list/ListRowView.kt
```diff
@@ -315,17 +316,19 @@ setHeadlineLayout(
...
 
-    private fun finishInit(styledAttrs: TypedArray) {
-        styledAttrs.recycle()
+    private fun finishInit() {
         isViewInitialized = true
         recalculateContentDescription()
     }
 
     private fun recalculateContentDescription(newContentDescriptionInfo: ContentDescriptionInfo? = null) {
+        if (hasCustomContentDescription) return
```
haynlo, line 330: In case the implementation provides a custom contentDescription we don't wan't to override it wit the component's default one whenever there is a view change like Title, Subtitle, etc.

dagonco (approved):
🔝

pmartinbTEF (approved):
LGTM

aweell (approved):

PR #447, opened 2025-07-29
Add HighLightCardButtonSettings.invalidatePaddings ANDROID-16601

jeprubio:
Add the possibility to invalidate the button paddings on HighlightedCard.

- Add the possibility to invalidate the button paddings on HightlithtedCard

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

- [ ] 🖼️ Screenshots/Videos

Tested by publishing the mistica lib locally:

| Before      | After |


jeprubio:
This works for our specific case but still HighlightedCard is using a ButtonLink inside, that button has min width and has it's text centered, that would result in wrong layout if the text is small, not sure if it's worth doing this fix if it won't look right if the text is small. Closing for now.

PR #450, opened 2025-08-11
ANDROID-16187 Update targetSDK to API 36

pmartinbTEF:
Update target sdk to API 36. Current target SDK is 34 so it has been checked Android 15 implications too.

Review libs for 16kb compatibility
Check edge to edge behaviour of Catalog app
Check [changes in Configuration behaviour ](https://developer.android.com/about/versions/15/behavior-changes-15#stable-configuration)
Review official documentation of [Android 15](https://developer.android.com/about/versions/15/behavior-changes-15) and [Android 16](https://developer.android.com/about/versions/16/behavior-changes-16)

### ☑️ Checks
- [x] Tested with dark mode.
- [x] Tested with API 24.

Check catalog app is working normally

### 🤖 Copilot review
@copilot Please review this PR for potential breaking changes in the public API **within the `library` module only**.

Focus on:
- Enum value removals.
- Removed or renamed public classes, interfaces, enums, methods, properties, or constants.
- Changes in method signatures (parameters, return types).
- Visibility changes (e.g., public → protected/private/package-private).
- Removal of public annotations or change of annotation values (e.g., @JvmOverloads, @Deprecated).
- Behavioral changes that keep the signature but alter the contract (e.g., new exceptions thrown, different return values for same input).

Review all modified files in the `library` module and comment on each breaking change.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -97,11 +97,12 @@ dependencies {
     implementation "com.github.skydoves:balloon-compose:1.5.3"
 
     implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.6.1"
+    implementation 'androidx.window:window:1.4.0'
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation "androidx.compose.ui:ui-test-junit4:$compose_ui_version"
     testImplementation "androidx.compose.ui:ui-test-manifest:$compose_ui_version"
-    testImplementation 'org.robolectric:robolectric:4.11.1'
+    testImplementation 'org.robolectric:robolectric:4.14'
     testImplementation "io.github.takahirom.roborazzi:roborazzi:$roborazzi_version"
     testImplementation "io.github.takahirom.roborazzi:roborazzi-compose:$roborazzi_version"
     testImplementation "org.mockito.kotlin:mockito-kotlin:4.0.0"
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -103,7 +103,7 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     testImplementation "androidx.compose.ui:ui-test-junit4:$compose_ui_version"
     testImplementation "androidx.compose.ui:ui-test-manifest:$compose_ui_version"
-    testImplementation 'org.robolectric:robolectric:4.14'
+    testImplementation 'org.robolectric:robolectric:4.15.1'
     testImplementation "io.github.takahirom.roborazzi:roborazzi:$roborazzi_version"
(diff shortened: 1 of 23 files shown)


library/build.gradle
```diff
@@ -58,6 +58,7 @@ android {
             includeAndroidResources = true
             all {
                 systemProperty 'robolectric.graphicsMode', 'NATIVE'
+                systemProperty 'robolectric.pixelCopyRenderMode', 'software'
```
pmartinbTEF, line 61: To avoid unwanted shadows in compose screenshots

library/build.gradle
```diff
@@ -97,11 +98,12 @@ dependencies {
     implementation "com.github.skydoves:balloon-compose:1.5.3"
 
     implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.6.1"
+    implementation 'androidx.window:window:1.4.0'
```
pmartinbTEF, line 101: The recommended way to get the screen width

library/src/main/java/com/telefonica/mistica/compose/tabs/Tabs.kt
```diff
@@ -61,7 +64,14 @@ fun Tabs(
     tabs: List<Tab>,
     onSelectedTabChanged: (index: Int) -> Unit,
 ) {
-    val isTablet = LocalConfiguration.current.screenWidthDp > 768
+
+    val context = LocalContext.current
+    val windowMetrics = WindowMetricsCalculator.getOrCreate().computeCurrentWindowMetrics(context as Activity)
```
pmartinbTEF, line 69: Recommended way to get screen width

library/src/main/java/com/telefonica/mistica/list/ListRowViewWithCheckBox.kt
```diff
@@ -37,7 +37,6 @@ class ListRowViewWithCheckBox @JvmOverloads constructor(
     fun changeCheckBoxState(newState: Boolean? = null) {
         checkBoxCompat.isChecked = newState ?: !checkBoxCompat.isChecked
         ViewCompat.setStateDescription(this, ViewCompat.getStateDescription(checkBoxCompat))
-        this@ListRowViewWithCheckBox.announceForAccessibility(ViewCompat.getStateDescription(this))
```
pmartinbTEF, line 40: Not necessary. Same behaviour as ListRowItemWithSwitch

jeprubio (approved):
Nice, I would check if we can place the Config inside the ScreenshotTests class

haynlo (approved):
LGTM!

